PR #1, opened 2016-12-06
Added functionality to set volume level.

soldag:
This adds method `set_volume` to explicitly setting the master volume of the AVR.


ol-iver (approved):
Looking good, I'll include that to the next version.
Thank you 👍

PR #2, opened 2016-12-15
Add moc/test xml from my AVR 1912

glance-:
These shows quite a bit of difference between the devices.


ol-iver:
Thank you for the files, it's good to be able to test with other receivers 👍 
I'll check the differences.

PR #5, opened 2017-03-12
use on/off states for mute settings

shapiromatron:
In order to be consistent with denon XML, use on/off state for `_mute` parameter instead of python booleans


ol-iver:
thank you for the fix :-)

PR #10, opened 2017-05-11
Try to detect error string in body

glance-:
The AVR-1912 returns 200 OK on POST to /goform/AppCommand.xml, but with
a error in the body.

Before this patch, that error was missed and a empty mapping for
renamed_sources and deleted_sources was returned, togeather with a
status of True, so _get_renamed_deleted_sources was never run.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -452,6 +452,9 @@ class DenonAVR(object):
         # Buffered XML not needed anymore: close
         body.close()
 
+        if 'Access Error: Data follows' in res:
+            return (renamed_sources, deleted_sources, False)
+
         try:
             # Return XML ElementTree
             root = ET.fromstring(res)


trahnel:
This is great! I have a Marantz SR5006 and debugged the `No mapping for audio source X` error I had and this was the source of the bug. Looking forward to get this merged.

ol-iver:
hey @glance- ,
thank you for the fix. I'll include that to my next version :)
Could you please send me an example of your AppCommand.xml for my local tests? You can use the small script attached, which will store the result to AppCommand_saved.xml in the same directory:
`python denon_appcommand_output.py --host 192.168.0.250`
[denon_appcommand_output.zip](https://github.com/scarface-4711/denonavr/files/983101/denon_appcommand_output.zip)

glance-:
```
$ python3 denon_appcommand_output.py --host 192.168.1.10
HTTP Status Code: 200
$ cat AppCommand_saved.xml 
<html><head><title>Document Error: Data follows</title></head>
		<body><h2>Access Error: Data follows</h2>
		<p>Form AppCommand.xml is not defined</p></body></html>

$
```

glance-:
[AppCommand_saved.zip](https://github.com/scarface-4711/denonavr/files/1019750/AppCommand_saved.zip)

ol-iver:
thanks :)
added the new version to homeassistant
https://github.com/home-assistant/home-assistant/pull/7779

PR #11, opened 2017-05-22
Add Marantz SSDP discovery

p3dda:
The denonavr Library also works for Marantz AVR, since they share the same XML API (at least as far as I know). I have tested this with with my Marantz NR 1604 and it works without problems.

I have added a small patch to enable automatic device discovery via SSDP for both, Denon and Marantz receivers.


ol-iver:
hey @p3dda,
cool, I will include that to the next version :)
cheers,
Oliver

ol-iver:
now also added to homeassistant
https://github.com/home-assistant/home-assistant/pull/7779

PR #14, opened 2017-06-04
Marantz SR5008 also uses the same XML format as the Denon AVR-X

russell:
When a Marantz SR5008 is detected, switch to the AVR-X access mode.


ol-iver:
thanks for your updates and your test files :)

PR #16, opened 2017-06-04
Add xml files for Marantz M-CR603

bmesuere:
When running, the following error is shown:
```
>>> d = denonavr.DenonAVR("192.168.0.3")
No mapping for audio source TUNER
Missing status information from XML for: FriendlyName
```


ol-iver:
I'll fix that in the next version, thank you :)

tests/xml/M-CR603-formMainZone_MainZoneXml.xml
```diff
@@ -0,0 +1,108 @@
...
+<value></value>
+<value></value>
+<value></value>
+<value></value>
+<value></value>
+<value></value>
+<value></value>
+</RenameSource>
+<RenameZone><value></value></RenameZone>
+<SourceDelete>
```
scarface-4711: @bmesuere: according to this SourceDelete section you deleted each input source on your receiver. The error message with the missing mapping for the TUNER is raised, because it was deleted.
Please check if you deleted (deactivated) all sources in your web interface or if this is a bug

tests/xml/M-CR603-formMainZone_MainZoneXml.xml
```diff
@@ -0,0 +1,108 @@
...
+<value>DEL</value>
```
ol-iver, line 71: @bmesuere: according to this SourceDelete section you deleted each input source on your receiver. The error message with the missing mapping for the TUNER is raised, because it was deleted.
Please check if you deleted (deactivated) all sources in your web interface or if this is a bug

bmesuere: The only source I deactivated is last.fm. As far as I know, there is no way to disable other sources on the MCR603. I also don't think there is a web interface.

ol-iver:
Created an option to display the deleted input functions as well
If you instanciate the receiver providing a name and the option, you should not get the error anymore
```
d = denonavr.DenonAVR("192.168.0.3", "receiver living room", True)
or
d = denonavr.DenonAVR("192.168.0.3", name="receiver living room", show_all_inputs=True)
```
In homeassistant I created a new option show_all_sources to enable this.
See here: https://github.com/home-assistant/home-assistant/pull/7949

PR #18, opened 2017-06-04
Add Marantz NETWORK audio source

russell:
This fixes problems with home assistant not reporting the device as playing when playing internet radio.


ol-iver:
@russel: all done :) I'll create a new version with your patches soon

PR #19, opened 2017-06-26
Add SR5009 to AVR-X Pattern

philk:
SR5009 has the exact same interface to the SR5008


ol-iver:
Thanks :)

ol-iver:
I'll include that to version 0.49 of homeassistant. Version 0.48 seems to be already packaged.

PR #27, opened 2017-07-18
Treat Marantz SR6008 as AVR-X device

bjw-s:
Similar to the change to include Marantz SR5006 as AVR-X device, also include Marantz SR6008. This prevents "No mapping for source MPLAY" errors.


ol-iver:
thanks :)

PR #29, opened 2017-08-21
Treat Marantz SR6009 and SR6010 as AVR-X device

marijngiesen:
This fixes source mapping errors in Home Assistant for Marantz SR6009 and SR6010

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -20,7 +20,7 @@ import requests
 
 _LOGGER = logging.getLogger("DenonAVR")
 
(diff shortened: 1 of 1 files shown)


kylehendricks:
I believe I am having the exact same issue with my Denon AVR-S920W.  I hardcoded `_avr_x` to true and everything started working.

It's starting to seem like we either need to start keeping track of which models need which input selection method or a different detection method.

kylehendricks:
Could this all be simplified to checking if `Deviceinfo.xml` exists?

Or possibly if `CommApiVers` > 0210

ol-iver:
@kylehendricks: I just added an additional criteria for CommApiVers. `CommApiVers > 0210` won't work as there is at least one Marantz receiver with version 0220 which does not work in the AVR-X mode. Thus I took 0210 and 0300 for now.

PR #31, opened 2017-09-17
Add MANIFEST.in so that tests are included in source distributions

sbraz:
Hi,
I checked the source and I don't see where pytest-* is required. I also added a MANIFEST.in file because the PyPI distributions currently do not contain the tests. Adding them makes it easier for distributions that rely on PyPI download sources and run tests.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,5 @@
+exclude .*
+exclude MANIFEST.in
+graft tests
+global-exclude __pycache__
+global-exclude *.py[co]


ol-iver:
Thanks.
The pytest* requirements could be removed as well. I'll check that.

ol-iver:
pytest is required by tox. See line 12 in [tox.ini](https://github.com/scarface-4711/denonavr/blob/master/tox.ini)

sbraz:
Yes, I noticed later and I fixed the PR accordingly :)

ol-iver:
this explains why I did not find any changes in the test requirements :)

PR #63, opened 2018-09-03
Added SoundModeRaw Support for Denon AVR 2016

berrywhite96:
The Denon AVR 2016 request also an other "AppCommand.xml", which wasnt included in the master branch. 
I added the get_app_command_0300_xml() function, for this file which can be extended for other attributes.
Unfortunately I didnt find out how to get the sound mode from different zones, it only request the sound mode for zone 1.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -64,6 +64,7 @@ ALBUM_COVERS_URL = "http://{host}:{port}/NetAudio/art.asp-jpg?{time}"
 
 # General URLs
 APPCOMMAND_URL = "/goform/AppCommand.xml"
+APPCOMMAND_0300_URL = "/goform/AppCommand0300.xml"
 DEVICEINFO_URL = "/goform/Deviceinfo.xml"
 NETAUDIOSTATUS_URL = "/goform/formNetAudio_StatusXml.xml"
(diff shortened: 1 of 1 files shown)


andrewsayre:
Sound mode on 2016 AVRs is also available through the AppCommand API using the GetSurroundModeStatus command.  This can be added to the existing _update_avr_2016 function so you don't need an additional call.  Additionally, you need to patch _get_support_sound_mode so it properly sets the value to true for 2016 AVRs.  Check out PR #65 where I implemented this functionality, in addition to All Zone Stereo support.

ol-iver:
@berrywhite96: That sounds like a good idea. Could you do that?

> 
> 
> Sound mode on 2016 AVRs is also available through the AppCommand API using the GetSurroundModeStatus command. This can be added to the existing _update_avr_2016 function so you don't need an additional call. Additionally, you need to patch _get_support_sound_mode so it properly sets the value to true for 2016 AVRs. Check out PR #65 where I implemented this functionality, in addition to All Zone Stereo support.

starkillerOG:
@scarface-4711 since sound mode support has in the meantime already been integrated for AVR-X-2016 type trough the AppCommand API (as mentioned in the post above), I think this PR can be closed.

@scarface-4711 can you close this PR?

PR #66, opened 2018-11-02
fix sound mode MULTI CH IN

starkillerOG:
I wrongfully put MULTI CH IN under MC STEREO, but it actually is a surround option that belongs under DTS SURROUND or DOLBY DIGITAL. (the marantz is put to MULTI CH IN if you select eather DTS SURROUND or DOLBY DIGITAL, and is put to MCH STEREO if you select MCH STEREO, when the input is a multi channel signal)


ol-iver:
Thanks
Included your changes into [HA](https://github.com/home-assistant/home-assistant/pull/18917).

starkillerOG:
Thank you @scarface-4711

PR #69, opened 2018-12-03
Add example how to use denonavr with gtk - simple volume control

PrzemoF:
I'm not sure if you're interested in including that kind of example how to use denonavr? I use it to control volume on DRA-100 - I couldn't find any linux desktop application doing that. Thanks for creating denonavr!


ol-iver:
@PrzemoF: I'm glad to hear that you like the library. Thanks
Your changes would intrduce a GTK dependency, which is not needed for its core functionality. We should not do this, to keep the library small.
Anyhow, keep on using it :smiley:

PR #76, opened 2019-02-20
Examples

robertschulze:
First of all thanks for providing this awesome package!
For my own purposes I have built a tiny infi.systray wrapper around denonavr to allow turning my Denon On/Off (plus doing rough adjustments to the volume) without searching for the remote (likely to be found in one of the children's rooms) or having to get up :-).
Instead of only using this myself, maybe it's worth to include this as part of an examples section?


robertschulze:
I believe that the example is better placed in a separate repo. I have updated it and created this new repo: [denonTray](https://github.com/robertschulze/denonTray)

PR #85, opened 2019-05-01
 Add Denon AVR-S series support

nellering:
Tested successfully on a Denon AVR-S930H.

Fixes issue scarface-4711#83.


ol-iver:
thanks, I'll include that into the next version 👍

PR #86, opened 2019-05-15
Get the sources list from Zone Status xml as a fallback.

### LINKED ISSUE #60: SR5006 Zone2 Source List Not Available
I started having a problem with my Zone2 sources several weeks ago, after updating Home Assistant. I'm not sure which version though, unfortunately...  

Zone2 sources do not appear in the **source_list** field, and I am unable to change Zone2 sources using the services tool in HA. All other Zone2 functions (power, volume, mute, etc.) still work, however. In Home Assistant, the error log shows the following two errors:
`Input function list for Denon receiver at host 192.168.1.127 could not be updated.`
and 
`Receiver sources list empty. Please check if device is powered on.`
These error log entries appear about 6 times a minute, presumably because this is the update interval.

I ran the **denon_receiver_xml.py** and have attached my XML files for your reference: [xml_test_sr5006.zip](https://github.com/scarface-4711/denonavr/files/2275687/xml_test_sr5006.zip)

Are there any further tests I can run to try and determine what is causing the issue? The XML for Zone2 appears normal, and has the sources listed, so I'm not sure where else to look...

I'm not entirely sure whether this is an issue with denonavr, or whether the media_player implementation of Zone2 controls in HA has changed, so my apologies in advance if I'm not posting this issue in the correct repo.

notgwj:
I think this fixes #60.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -791,9 +791,11 @@ class DenonAVR:
             if self._receiver_type == AVR_X.type:
                 root = self.get_status_xml(self._urls.status)
             # URL only available for Main Zone.
-            elif (self._receiver_type == AVR.type and
-                  self._urls.mainzone is not None):
+            elif (self._receiver_type == AVR.type):
+              if (self._urls.mainzone is not None):
                 root = self.get_status_xml(self._urls.mainzone)
+              else:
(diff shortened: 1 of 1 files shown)


ol-iver:
Sounds reasonable. Thanks 👍

PR #94, opened 2019-08-26
Explored the HTTP/XML commands being sent from app

JPHutchins:
In hopes of expanding the functionality of this library, specifically for use in Home Assistant, I have gathered some more info about the commands that Denon is receiving over HTTP.  

Note the AppCommand0300.xml endpoint that corresponds to the <cmd id="3"> tag.

I am particularly interested in creating a "night mode" macro as it is a nice thing to add to automation in Home Assistant.  

At first I was thinking about how to parse and generate these commands but since it is a closed API that they apparently don't want us to access it seems a waste of time.  It would be relatively easy to create a folder called "commands" and place in it XML files corresponding to specific commands or potentially plaintext with metadata (in JSON or YAML) above the XML that informs a python function how the variables within the XML should be handled.

Also of interest perhaps to those that know more is the HTTP SUBSCRIBE method that appears in the first handshake between iPhone and AVR.  I was not able to replicate responses in Postman so i didn't grab those examples.

### DIFF
--- a/XML_data_dump.txt
+++ b/XML_data_dump.txt
@@ -1,0 +1,463 @@
+Network captures from Denon X1500H 25 Aug 2019
+Commands sent with Denon 2016 AVR Remote Version 3.2.0
+on an iPhone 6
+
+The endpoint is /goform/AppCommand.xml unless the entry is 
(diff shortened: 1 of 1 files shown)


ol-iver:
thanks :)

PR #103, opened 2019-10-16
Import all 3313 IR commands; test a mute toggle.

JPHutchins:
It is speculation, but so far we know that these commands work for AVR X4400H (2017), AVR X1500H (2018), and the source document is for AVR 3313 (2012), so they may be the most compatible set of codes that we have.

It would not be my intention to write many methods such as the `mute_toggle` in this PR, but rather to make available the general purpose method `send_post_ir` that allows the user to send generic IR commands.  This allows the implementer of the library, including Home Assistant, to add specific IR commands to suit their needs without bogging down this library with more API variations than it already has.  To this end, `import denonavr.ir_code as ir` gives the user or developer access to a library that may allow them to achieve their specific goals.

Eventually these codes may become outdated and we can continue with the XML command API or some other variation.

Thank you to @bendavid for taking the time to verify that these codes work via the `iPhoneAppDirect.xml` endpoint and providing a link to the exhaustive list!

Cheers,
J.P.


ol-iver:
Thanks for implementing this @JPHutchins . 
I'm not able to test it, because apparently my Denon does not support those commands. 
Could you please remove the PDF File from docs folder, because it looks like a part of a Denon manual and should not been published here.

Edit: my Denon AVR-X4100W is supporting the commands, but the HTTP request can be handled correctly only when using GET instead of POST. Please see comments below.

denonavr/ir_code.py
```diff
@@ -0,0 +1,507 @@
...
+RETURN = 'RCKSK0740073'
+SHIFT_A = 'RCKSK0740096'
+SHIFT_B = 'RCKSK0740097'
+SHIFT_C = 'RCKSK0740098'
+SHIFT_D = 'RCKSK0740099'
+SHIFT_E = 'RCKSK0740100'
+SHIFT_F = 'RCKSK0740101'
+SHIFT_G = 'RCKSK0740102'
+BAND_FM = 'RCKSK0740153'
+BAND_AM = 'RCKSK0740154'
```
scarface-4711: Please provide a valid docstring for this .py file and convert the constants to upper case only, that they fit the python naming conventions.

doc/read_avr3313_pdf.py
```diff
@@ -0,0 +1,21 @@
...
+            name = name.replace('__', '_')
+            name = name.replace('▼', 'DOWN')
+            name = name.replace('▲', 'UP')
+            name = name.replace('.', '_')
+            if name[-1] == '_':
+                name = name[:-1]
+
+            code = line[:12]
+            w.write(f'{name} = \'{code}\'\n')
+        line = f.readline()
```
scarface-4711: same as for `ir_code.py`, please add a docstring and convert the `name` to upper case

denonavr/ir_code.py
```diff
@@ -0,0 +1,507 @@
+ALL_POWER_ON_OFF = 'RCKSK0410001'
```
ol-iver, line 1: Please provide a valid docstring for this file

denonavr/ir_code.py
```diff
@@ -0,0 +1,507 @@
...
+CURSOR_UP = 'RCKSK0410027'
+CURSOR_DOWN = 'RCKSK0410028'
+CURSOR_LEFT = 'RCKSK0410029'
+CURSOR_RIGHT = 'RCKSK0410030'
+ENTER = 'RCKSK0410031'
+RETURN = 'RCKSK0410034'
+SETUP_MENU = 'RCKSK0410035'
+MEMORY = 'RCKSK0410036'
+UPGRADE_ID_ON = 'RCKSK0410042'
+Bilingual_Mode_toggle__JP_only = 'RCKSK0410045'
```
ol-iver, line 30: Please convert all constant names to upper case that they fit to the python style guide

doc/read_avr3313_pdf.py
```diff
@@ -0,0 +1,21 @@
+with open('avr3313.txt', 'r') as f, open('ir_code.py', 'w') as w:
```
ol-iver, line 1: Please add a docstring here too

doc/read_avr3313_pdf.py
```diff
@@ -0,0 +1,21 @@
...
+            name = name.replace('-', '_')
+            name = name.replace('+', '_')
+            name = name.replace('(', '_')
+            name = name.replace(')', '_')
+            name = name.replace('/', '_')
+            name = name.replace(':', '_')
+            name = name.replace('__', '_')
+            name = name.replace('▼', 'DOWN')
+            name = name.replace('▲', 'UP')
+            name = name.replace('.', '_')
```
ol-iver, line 15: please convert `name` to upper case that constant names fit the python style guide

ol-iver (changes requested):

denonavr/denonavr.py
```diff
@@ -394,6 +396,18 @@ def send_post_command(self, command, body):
                 "end point %s"), self._host, res.status_code, command)
             return False
 
+    def send_post_ir(self, command):
+        """Send IR command via HTTP post to receiver."""
+        res = requests.post("http://{host}:{port}{uri}{command}".format(
```
ol-iver, line 401: Could you change this to a GET call?
I know it does not sound right, but I'm getting a _Read timed out_ error on HTTP POST, though the command was sent successfully. Using a GET request the command is still working, but there is not _Read timed out_ error

denonavr/denonavr.py
```diff
@@ -394,6 +396,18 @@ def send_post_command(self, command, body):
                 "end point %s"), self._host, res.status_code, command)
             return False
 
+    def send_post_ir(self, command):
```
ol-iver, line 399: Please rename to `send_get_ir` if the change below works

PR #106, opened 2019-11-30
Add some sound modes

starkillerOG:
Add 3 sound modes to solve 2 outstanding issues
will solve: https://github.com/scarface-4711/denonavr/issues/105 and https://github.com/scarface-4711/denonavr/issues/99


ol-iver:
thanks again :blush:

Issue #78, opened 2019-03-11
Sound mode ALL_ZONE_STEREO to sticky


DanskerUS:
Hi, thanks for all the work the denonavr module.

Using the sound modes: after ALL_ZONE_STEREO I can no longer switch to the other sound modes. 

It is a special Denon feature, that ALL_ZONE_STEREO must be OFF, or you cannot set any of the other sound modes.

Just like in Harmony Hub, there needs to be a ALL_ZONE_STEREO_OFF option. Or, call ALL_ZONE_STEREO_OFF internally before setting the new sound mode. 

Can help with testing, have: X1000, X2100 and X3400

Thank you again !

andrewsayre:
Hmm...  There is code to turn it off before switching to another sound mode: https://github.com/scarface-4711/denonavr/blob/master/denonavr/denonavr.py#L1453-L1455

Is it sticking in all circumstances?  I could see the mode not being reported back if the device is power-cycled and the library doesn't know it's active anymore.

DanskerUS:
Andrew,

Thank you very much for your reply!

Did get X3400H added to HA along with X1000 and X2100W. Did some testing on all of them.

First, I did test for the point you make about HA and Denon device getting out of sync. In most of my use cases, a mix of voice, HA and OEM IR remotes are used. Yes, turning the Denon off/on using the OEM IR remote results in the Denon's sound mode reported as "sound mode" or blank. I can set it from HA, and it works. The on/off, volume and source is being synced/pooled, but apparently sound mode is not.

For all three, a main zone Off/On always clears ALL_ZONE_STEREO mode.

There is something else going on:

X3400H, X2100W and X1000: When ALL_ZONE_STEREO is on, the Off command to either main OR zone2, turns both off. If so, the ALL_ZONE_STEREO could/should be set to Off before the main or zone2 Off is sent; that way, basic on/off works the same for all sound modes.

Also:

On the X3400H and X2100W: Once the ALL_ZONE_STEREO is ON (by either IR or HA), then HA fails to set any other sound mode, Only way to clear ALL_ZONE_STEREO on mode, it to power cycle. The Denon OEM IR remote can turn the sound mode off.

On the X1000: sounds modes can be set as expected (code to turn off ALL_ZONE_STEREO first works!).

And:

Source selection is a bit different on the three Denons if ALL_ZONE_STEREO is on, as follows:

X1000: main zone can be set on anything, zone2 two on "Source" (same as main zone), or RCA audio inputs.

[...]

PR #107, opened 2019-12-14
Fix bug preventing All Zone Stereo from being disabled automatically

JeremyKennedy:
Likely related to issue #78. My receiver reports `self._sound_mode_raw` as `All Zone Stereo`, however your `if` in `set_sound_mode:1466` expects it to be uppercase - `ALL ZONE STEREO`. This change simply converts it to uppercase within the `if` statement, allowing All Zone Stereo to be disabled when necessary. I'm guessing that some receivers report it all caps, but mine (X1500H) uses title case.

It may be better to instead set  `self._sound_mode_raw` to uppercase automatically, rather than only in the `if`, however I didn't want to mess with too much.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -1463,7 +1463,7 @@ class DenonAVR:
                 return True
             else:
                 return False
-        if self._sound_mode_raw == ALL_ZONE_STEREO:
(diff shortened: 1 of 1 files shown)


ol-iver:
Let's start with this, thank you :blush:

PR #111, opened 2020-01-02
alternative fix for #107

starkillerOG:
I think using self.sound_mode is better than self._sound_mode_raw.upper(), because the whole point of self.sound_mode is that it is matched to a single value rather than the diffrent raw_sound_modes reported by the receiver (accross diffrent models).

Note that I have not tested this, because my receiver does not support ALL_ZONE_STEREO.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -1465,7 +1465,7 @@ class DenonAVR:
                 return True
             else:
                 return False
-        if self._sound_mode_raw.upper() == ALL_ZONE_STEREO:
+        if self.sound_mode == ALL_ZONE_STEREO:
             if not self._set_all_zone_stereo(False):
                 return False
         # For selection of sound mode other names then at receiving sound modes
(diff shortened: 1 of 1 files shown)


starkillerOG:
I think this could be a more future prove fix than #107

starkillerOG:
@JeremyKennedy and @scarface-4711 do you agree?

ol-iver:
@starkillerOG There should be a mapping from `All Zone Stereo` to `ALL_ZONE_STEREO` in `SOUND_MODE_MAPPING` for this to work, right?

starkillerOG:
@scarface-4711 you are right, but there is in line 1526:
`            sound_mode = self._sm_match_dict[sound_mode_raw.upper()]`
The .upper() already takes care of it inside the match_sound_mode function.

The ALL_ZONE_STEREO  is a constant with value 'ALL ZONE STEREO', I do not like it in the matching dict so I will also change that to just be the typed out version for uniformity.

ol-iver:
ok got your point 😄

JeremyKennedy:
thanks @starkillerOG and @scarface-4711! glad to have this fixed :)

PR #119, opened 2020-02-20
AVR_X type: try update_avr_2016 to fix Spotify identify

starkillerOG:
I noticed that my Marantz SR5008 reported "Network" as source while playing form spotify instead of reporting "Spotify", I have been searching for a way to fix this for a long time.
Recently I discovered that the SR5008 supports Appcommand and therefore the update_avr_2016 function works fine for my SR5008. The update_avr_2016 does report "Spotify Connect" when playing from spotify. (needed to add source mapping "Spotify Connect"-->"SpotifyConnect" and now state is correctly reported for spotify.

Besides this adds a third redandancy update for AVR-X device types.


starkillerOG:
@scarface-4711 Could you look at this and merge it?

ol-iver:
seems to run smoothly, thx 😃

PR #125, opened 2020-03-22
DEVICEINFO_COMMAPI_PATTERN pattern update

daha76:
Update DEVICEINFO_COMMAPI_PATTERN to match latest firmware update for NR1609 and probably other AV receivers.


ol-iver:
tested it with your xml files and it seems to work. Thanks 😄

daha76:
> tested it with your xml files and it seems to work. Thanks 😄

No Problem ;-)

PR #127, opened 2020-03-29
Adding support for Python 2.7.  Specific changes:

Ben-Mathews:
- Explicitly inheirting DenonAVR from object (needed to support super() call)
 - Manual list clears (since clear() does not exist for lists in 2.7)
 - Python 2.7 super() call fix
 - Importing from urlparse instead of urllib.parse
 - Defining FileNotFoundError exception (since this doesn't exist in 2.7)
 - using io.open() because vanilla open() does not support encoding keyword


ol-iver:
Please see my comment in the related issue: https://github.com/scarface-4711/denonavr/issues/128#issuecomment-630286560

PR #129, opened 2020-04-05
fix FutureWarning

starkillerOG:
When looking at the log of HomeAssistant I came occros this:
`/srv/homeassistant/lib/python3.7/site-packages/denonavr/denonavr.py:1031: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.`
This fixes that

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -1028,7 +1028,7 @@ class DenonAVR:
             receiver_sources = {}
             # Source determination from XML
             favorites = root.find(".//FavoriteStation")
-            if len(favorites):
+            if favorites is not None:
                 for child in favorites:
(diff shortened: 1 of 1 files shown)


denonavr/denonavr.py
```diff
@@ -1028,7 +1028,7 @@ def _get_receiver_sources(self):
             receiver_sources = {}
             # Source determination from XML
             favorites = root.find(".//FavoriteStation")
-            if favorites:
+            if len(favorites):
```
ol-iver: imo this looks awkward , `If favorites is not None` is better 😃

starkillerOG: just update it

PR #132, opened 2020-04-30
Add device info such as serial_number

starkillerOG:
This is needed to construct a unique ID that is needed for the Config Flow of HomeAssistant


starkillerOG:
This PR also fixes the denonavr.discover()
That has never worked for me on any of my computers.
It turns out the fix is to bind the ssdp discovery socket to the local ip adress.
Now it works like a charm on both windows, virual ubuntu on windows and linux computers.

starkillerOG:
@scarface-4711 Could you please look at this PR (and also the less important other PRs of me)?
I am working on a big HomeAssistant PR that will implement Config Flow for the DenonAVR platform. That includes unique IDs and that is why this PR is needed. Furthermore the discovery of denon receivers has been imporved and I am busy on implementing the new config flow ssdp discovery of homeassistant.

Could you merge this PR and push to a new version of denonavr?

starkillerOG:
The HomeAssistant branch that implements Config Flow is here: https://github.com/starkillerOG/home-assistant/tree/patch-23

starkillerOG:
@scarface-4711 the HomeAssistant PR is here: https://github.com/home-assistant/core/pull/35129
Could you please merge this as soon as possible, then I can finisch the HomeAssistant PR?

denonavr/ssdp.py
```diff
@@ -49,11 +50,13 @@ def ssdp_request(ssdp_st, ssdp_mx=SSDP_MX):
         'HOST: {}:{}'.format(*SSDP_TARGET),
         '', '']).encode('utf-8')
 
-def get_local_ip(sock):
-    ip_list = socket.gethostbyname_ex(socket.gethostname())[2]
-    for ip in ip_list:
-        if ip.startswith("192."):
-            return ip
+def get_local_ip():
```
JPHutchins: It is not necessarily the case that local addresses start with 192 - they can really start with 10 172 or 192 - we would like to filter out self-assigned, those that start with 169.
Further, this (and the older) code will give me 127.0.0.1 (local address) not an external IP.  I have done some digging and this function works for me to get my external IP (_happens_ to be 192):
```python
import socket
def get_ip():
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        # doesn't even have to be reachable
        s.connect(('10.255.255.255', 1))
        IP = s.getsockname()[0]
    except:
        IP = '127.0.0.1'
    finally:
        s.close()
    return IP
```
I found this here after a lot of fails with socket.gethostname() https://stackoverflow.com/questions/166506/finding-local-ip-addresses-using-pythons-stdlib

JPHutchins (approved):

denonavr/denonavr.py
```diff
@@ -29,6 +31,11 @@
 AVR_X = ReceiverType(type="avr-x", port=80)
 AVR_X_2016 = ReceiverType(type="avr-x-2016", port=8080)
 
+DiscriptionType = namedtuple('DiscriptionType', ["port", "url"])
```
JPHutchins: Spelling: should this be "DescriptionType" and otherwise change discription to description?

denonavr/ssdp.py
starkillerOG: If I do 127.0.0.1 it does not work for me, I do not get any discovered devices.
If I run the code of this PR I get an IP of 192.168.1.xxx which is the IP adress of the computer running the code in my local network, then I do discover my receiver.
If I run the code you perposed I get an IP of 10.xx.xx.xx and I do not discover my receiver using that code....
So this will not work (at least not on my system), but apperently my code does not work on your system.

I will rework to use all IP's discovered just like is done in netdisco library.

starkillerOG: Note that my system is running a VPN for internet connections, but not for local traffic.

denonavr/denonavr.py
starkillerOG: thanks for pointing this out, just fixed it

denonavr/ssdp.py
starkillerOG: The new PR schould fix it for you, it works on my system.

Does it now also work on your system?

starkillerOG:
@JPHutchins would you be willing to look at the test that is failing on the HomeAssistant PR?
I have a hard time getting that fixed and frankly i am not that good at the whole testing using those mock classes.

I will start working on the tests of the config flow that are needed.

denonavr/denonavr.py
```diff
@@ -419,6 +433,26 @@ def send_post_command(self, command, body):
...
+        url = "http://{host}:{port}{command}".format(
+            host=self._host, port=DESCRIPTION_URL[self._receiver_type].port,
+            command=DESCRIPTION_URL[self._receiver_type].url)
+        
+        device_info = evaluate_scpd_xml(url)
+        
+        if device_info == False:
+            self._manufacturer = "Denon"
+            self._model_name = "Unknown"
+            self._serial_number = self._host
```
bieniu: I think the IP address shouldn't be used in Home Assistant as `unique_id` because it can changed. The MAC address will be a better choice.

starkillerOG: I do not use the IP address, only if getting the serialNumber fails I use the IP address as a backup, but I also print an error in the log about it, so users that experiance problems getting the serialNumber will see that error and hopefully report it such that it can be fixed. The IP adress is then only used in the meantime for these users to be able to still setup the denonavr platform.

I could also just simply remove this "alternative" behaviour...

I do not know the MAC address of the receiver. It is specified in the device_info.xml, but I do not have a way of retrieving it from the appcommand so on newer avr-x-2016 models I have no way of getting the MAC, therefore implementing MAC at this point would break it for the avr-x-2016 models.

Unless someone could figure out how to get the MAC form the appcommand?

starkillerOG:
@scarface-4711 could you please merge this?
It has been looked at by 2 other persons and the ConfigFlow PR is all ready, just need this to be merged and a version push.

denonavr/denonavr.py
bieniu: But the `serial_number` property shouldn't contain the IP address, even if the real serial number cannot be obtained. Maybe the library should return `None` when there is no serial number and the HA integration should get the MAC address via `getmac` package?

starkillerOG: I removed the ip adress being used as serial_number, I will also make a change to the homeassistant PR

ol-iver:
@starkillerOG thanks for your work. I'll have a look into it now 😄

starkillerOG:
Thank you, that would be greath!

denonavr/denonavr.py
```diff
@@ -307,6 +318,9 @@ def __init__(self, host, name=None, show_all_inputs=False, timeout=2.0,
 
         # Determine receiver type and input functions
         self._update_input_func_list()
+        
+        # Determine device info
+        self.get_device_info()
```
ol-iver, line 323: What happens if the device is offline at this point in time? Maybe we could check if there are device infos on update.

denonavr/denonavr.py
```diff
@@ -419,6 +433,26 @@ def send_post_command(self, command, body):
...
 
+    def get_device_info(self):
+        """Get device information."""
+        url = "http://{host}:{port}{command}".format(
+            host=self._host, port=DESCRIPTION_URL[self._receiver_type].port,
+            command=DESCRIPTION_URL[self._receiver_type].url)
+        
+        device_info = evaluate_scpd_xml(url)
+        
+        if device_info == False:
```
ol-iver: `device_info is False` is nicer

denonavr/ssdp.py
```diff
@@ -13,51 +13,51 @@
 import xml.etree.ElementTree as ET
 from urllib.parse import urlparse
 import requests
+import ifaddr
```
ol-iver, line 16: This is a 3rd party package, so it should be included into setup.py

ol-iver (changes requested):
Little remarks

denonavr/ssdp.py
```diff
@@ -13,51 +13,51 @@
...
+        'MAN: "ssdp:discover"',
+        'HOST: {}:{}'.format(*SSDP_TARGET),
+        '', '']).encode('utf-8')
+
+def get_local_ips():
+    adapters = ifaddr.get_adapters()
+    ips = []
+    for adapter in adapters:
+        for ip in adapter.ips:
+            if type(ip.ip)==str:
```
ol-iver: `if isinstance(ip.ip, str)` please

scarface-4711: checking this is not neccessary anyhow. What else should it be?

denonavr/denonavr.py
starkillerOG: For me personally the receiver is always reachable even if it is off, but I did not think of other models that might not respond if they are in standby.

Will add a check in the update

denonavr/ssdp.py
starkillerOG: > checking this is not neccessary anyhow. What else should it be?

adapter.ips includes both Ipv4 and Ipv6 addresses, the Ipv6 addresses are tuples instead of strings and the code can't deal with Ipv6, so I filter those out.

denonavr/ssdp.py
starkillerOG: thanks!
I was already worried about this, I did not know how to make sure someone has this module installed.

Has now been fixed

denonavr/denonavr.py
starkillerOG: As it is now, if the receiver will not respond to the request you will get a ConnectionError, this is picked up by HomeAssistant and a ConfigEntryNotReady will be thrown. I think that means HomeAssistant will mark the entity as unavailable and will keep retrying untill it gets a connection and then setup the entry.

starkillerOG: So I don't think this is a problem.

The thing is that the way in which the Config Flow is made right now requires data from the receiver at startup like the mac adress and manufacturer etc. to setup the device entry.

I could store all that data in the config_registry at the moment the Config Flow is setup and reuse that data at a restart of HomeAssistant, but that will require some more coding and time.

I think almost all Denon and Marantz receivers always respond to requests even if they are in standby, so then this is no issue. Otherwise I am pretty sure the component will still be setup as soon as the receiver is turned on for the first time since a HomeAssistant restart.

ol-iver: Not everyone is keeping their receivers in standby all the time. Thus I would not assume that the device is running everytime when home-assistant is starting.
There are other places in the code where this seems to be not working all the time. I'll merge und fix this all together

ol-iver (approved):

PR #136, opened 2020-05-11
Fix discovery for AVR-X1600H

MarBra:
The upnp looks different for AVR-X1600H.
I have added a test to verify it for my model. Maybe we can add more tests for other devices.

### DIFF
--- a/denonavr/ssdp.py
+++ b/denonavr/ssdp.py
@@ -36,7 +36,10 @@ SCPD_SERIALNUMBER = "{xmlns}serialNumber".format(xmlns=SCPD_XMLNS)
 SCPD_FRIENDLYNAME = "{xmlns}friendlyName".format(xmlns=SCPD_XMLNS)
 SCPD_PRESENTATIONURL = "{xmlns}presentationURL".format(xmlns=SCPD_XMLNS)
 
-DEVICETYPE_DENON = "urn:schemas-upnp-org:device:MediaRenderer:1"
+SUPPORTED_DEVICETYPES = [
+    "urn:schemas-upnp-org:device:MediaRenderer:1",
+    "urn:schemas-denon-com:device:AiosDevice:1"
+    ]
 
 SUPPORTED_MANUFACTURERS = ["Denon", "DENON", "Marantz"]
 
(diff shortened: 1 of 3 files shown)


JPHutchins (commented):
Confirming successful discovery for Denon AVR-X1500H.  I would expect it to work for all receivers in the AVR-2016 class.  @starkillerOG does this help with the need for unique device ID in HA?  It is reporting my serial number.

starkillerOG:
Yes thank you @JPHutchins
The AVR-x-2016 uses a different URL, so it is nice to know it works for those receivers too.

starkillerOG:
@scarface-4711 looks good to me, could this be merged?

@JPHutchins, @MarBra is the Upnp for Heos speakers also:
`"urn:schemas-denon-com:device:AiosDevice:1"`?
Because I don't think that the denonavr integration works for Heos speakers, only for Denon and Marantz receivers that support Heos.
I do not want to wrongfully discover Heos speakers.....

MarBra:
> @scarface-4711 looks good to me, could this be merged?
> 
> @JPHutchins, @MarBra is the Upnp for Heos speakers also:
> `"urn:schemas-denon-com:device:AiosDevice:1"`?
> Because I don't think that the denonavr integration works for Heos speakers, only for Denon and Marantz receivers that support Heos.
> I do not want to wrongfully discover Heos speakers.....

Not sure, if it will also detect Heos speakers. I don't have one.

If you search for `serialNumber` in tests/xml/AVR-X1600H_upnp.xml you see that there are three "devices" with `serialNumber` tag. We can also use one of them to discover the device.
These `deviceTypes` are suitable:
- `urn:schemas-denon-com:device:AiosDevice:1` 
- `urn:schemas-denon-com:device:ACT-Denon:1`
- `urn:schemas-upnp-org:device:MediaServer:1`

starkillerOG:
I think `urn:schemas-denon-com:device:ACT-Denon:1` would be better to avoid discovering HEOS speakers, but I think `urn:schemas-denon-com:device:AiosDevice:1` would also be fine, we could always filter out Heos later if they cause problems

MarBra:
@starkillerOG I've chosen `urn:schemas-upnp-org:device:MediaServer:1` now, because `urn:schemas-denon-com:device:ACT-Denon:1` has a `ACT` prefix in `friendlyName`.

This could be easily changed, because we are looping over a `deviceList`, if one exists.

denonavr/ssdp.py
```diff
@@ -174,28 +178,44 @@ def evaluate_scpd_xml(url):
...
+                    in SUPPORTED_DEVICETYPES):
+                device_xml = root.find(SCPD_DEVICE)
+            elif root.find(SCPD_DEVICE).find(SCPD_DEVICELIST):
+                for dev in root.find(SCPD_DEVICE).find(SCPD_DEVICELIST):
+                    if (dev.find(SCPD_DEVICETYPE).text in SUPPORTED_DEVICETYPES
+                            and dev.find(SCPD_SERIALNUMBER) is not None):
+                        device_xml = dev
+                        break
+
+            if not device_xml:
```
ol-iver: This would create the same FutureWarning as solved here:
https://github.com/scarface-4711/denonavr/pull/129/commits/cd168d01d2725806a671852378c8bfe47d12795d#diff-18d32fba6089c732c810ffa526db114cR1031

denonavr/ssdp.py
```diff
@@ -174,28 +178,44 @@ def evaluate_scpd_xml(url):
...
+                return False
+
+            if device_xml.find(SCPD_PRESENTATIONURL):
```
ol-iver: same FutureWarning thing here

denonavr/ssdp.py
MarBra: Done

denonavr/ssdp.py
MarBra: Done

JPHutchins (approved):

PR #137, opened 2020-05-17
Replace ifaddr with more prevalent netifaces

sbraz:
Hi,
I currently package denonavr for Gentoo and I think netifaces would be a better choice because it seems to be much more widely used:
* It had 27 times more PyPI downloads last month (https://pypistats.org/packages/netifaces vs https://pypistats.org/packages/ifaddr)
* it is in in about 3 times more repositories (https://repology.org/project/python:netifaces/versions vs https://repology.org/project/python:ifaddr/versions)
* it had a commit in November 2019 while ifaddr had its last commit in December 2018
* 13 packages from the Gentoo tree use netifaces, 1 uses ifaddr

@starkillerOG are you OK with that change since you were the one to introduce ifaddr? The current behaviour is the same as with ifaddr: only IPv4s are returned ; however this could easily be changed.


ol-iver:
Sounds reasonable, always go for maintained packages. I'll include that into the next version, if there are no objections from you @starkillerOG

starkillerOG:
Just checked the code that uses netifaces and it gives me the same results.
So I agree a better maintained packages is always good!

sbraz:
Thanks, can you release a new version with those changes? Otherwise, I'll backport the patch to 0.9.0.

ol-iver:
I will collect some changes to create version 0.9.1 today

ol-iver:
done

PR #138, opened 2020-05-23
Discover same device only once

MarBra:
My receiver (AVR-X1600H) will be discovered multiple times(13).

### DIFF
--- a/denonavr/ssdp.py
+++ b/denonavr/ssdp.py
@@ -81,7 +81,7 @@ def identify_denonavr_receivers():
             receiver = evaluate_scpd_xml(device["URL"])
         except requests.exceptions.RequestException:
             continue
-        if receiver:
+        if receiver and receiver not in receivers:
             receivers.append(receiver)
 
     return receivers


JPHutchins (approved):
Edit 2: I submitted a PR with the changes outlined below as an alternative to this commit.  https://github.com/scarface-4711/denonavr/pull/140

Tested and working with my 1500H.  It did have the same issue and this resolves it, good catch!

Edit: I changed my mind about this - I think it is better we catch the duplicates earlier and thereby avoid parsing them with the evaluate_scpd_xml function.  I tried a few things with map lambdas but then I realized that we are only using one value from the send_ssdp_broadcast() dict, so I propose simplifying that function instead.  Now it will simply return a set of unique ssdp URLs so we don't worry about parsing the same xml multiple times.

Let me know what you think about these ideas:

```python
def identify_denonavr_receivers():
    """
    Identify DenonAVR using SSDP and SCPD queries.

    Returns a list of dictionaries which includes all discovered Denon AVR
    devices with keys "host", "modelName", "friendlyName", "presentationURL".
    """
    # Sending SSDP broadcast message to get devices
    urls = send_ssdp_broadcast()

    # Check which responding device is a DenonAVR device and prepare output
    receivers = []
    for url in urls:
        try:                                                                                                                                                                                                                              receiver = evaluate_scpd_xml(url)
        except requests.exceptions.RequestException:
            continue
        if receiver:
            receivers.append(receiver)

    return receivers
```
Corresponding changes to the bottom of the send_ssdp_broadcast function:
```python
# Prepare output of responding devices
    urls = set()

    for entry in res:
        # Some string operations to get the receivers URL
        # which could be found between LOCATION and end of line of the response
        en_decoded = entry[0].decode("utf-8")
        # If location is not found, skip the entry
        try:
            url = (
                en_decoded[
                    en_decoded.lower().index(
                        "location:") + 10:en_decoded.index(
                            "\r\n", en_decoded.lower().index("location:"))]
            )
        except ValueError:
            continue

        urls.add(url)

    _LOGGER.debug("Following devices found: %s", urls)
    return urls
```

ol-iver:
Simplifying the broadcast function looks a bit better to me. Closed in favor of https://github.com/scarface-4711/denonavr/pull/140
Thanks @MarBra & @JPHutchins 😃

PR #142, opened 2020-06-07
fix getting receiver name for avr-x-2016

starkillerOG:


denonavr/denonavr.py
```diff
@@ -469,10 +469,7 @@ def ensure_configuration(self):
             if self.manufacturer is None and self.model_name is None:
                 executor.submit(self.get_device_info)
 
-            if self._receiver_type == AVR_X_2016.type:
-                executor.submit(self._get_zone_name)
-            else:
-                executor.submit(self._get_receiver_name)
+            executor.submit(self._get_receiver_name)
```
ol-iver, line 472: This removing functionality rather then a fix.
In case of multiple zone each zone would have the same name.

denonavr/denonavr.py
```diff
@@ -819,23 +816,45 @@ def _update_input_func_list(self):
...
-            except (ValueError, requests.exceptions.RequestException):
-                _LOGGER.warning("Receiver name could not be determined. "
-                                "Using standard name: Denon AVR.")
-                self._name = "Denon AVR"
-            else:
-                # Get the tags from this XML
-                name_tag = self._get_status_from_xml_tags(root, name_tag)
-                if name_tag:
+        if self._name is None:
+            if self._receiver_type == AVR_X_2016.type:
```
ol-iver, line 820: appcommand and status xml endpoints return exactly the same attribute. Thus there is no need to build two different cases.

ol-iver (commented):
This does not look like the right way.
There should be an attribute for the name of the device and the name of the media player instance. `self._name` is supposed to be the later one.

ol-iver:
Please see my comment. We should add multiple zone support to config flow instead.

starkillerOG:
@scarface-4711 I see your point.
However if using self._get_zone_name the entities will be called "Main Zone", "Zone2", "Zone3" which is also not ideal...

What about combining self._get_receiver_name and self._get_zone_name to get "{FriendlyName} Main Zone", "{Friendlyname} Zone2" and "{Friendlyname} Zone3"?

denonavr/denonavr.py
starkillerOG: @scarface-4711 the point is that the appcommand was not implemented and AVR_X_2016.type does not have a status xml endpoint anymore as far as I know.
Therefore this would add the ability to get the friendlyName for avr-x-2016 receivers, which is currently not possible as far as I can tell

denonavr/denonavr.py
starkillerOG: see my comment below

ol-iver:
Those are just the default names. Zones can be renamed on the config page of the receiver. My Main Zone is called "Wohnzimmer".

starkillerOG:
> 
> 
> Please see my comment. We should add multiple zone support to config flow instead.

multiple zones are supported and tested by the config flow

starkillerOG:
> 
> 
> Those are just the default names. Zones can be renamed on the config page of the receiver. My Main Zone is called "Wohnzimmer".

fair point, so maybe not change lines 473 to 475, but the rest is still valid.
The option to get the FriendlyName for avr-x-2016 receivers schould still be merged

denonavr/denonavr.py
starkillerOG: have reverted this in my fork

ol-iver:
But the method is not called anywhere for avr-x devices

ol-iver:
I checked how the zone names for the other device types are created. This is simply FriendlyName + Zone2 or Zone3. This zone stuff for avr-x-2016 was overkill. You are able to change the entity name in home-assistant anyway.
In the version 0.9.3 now every device shares the same naming convention.

PR #144, opened 2020-07-04
Add bass, treble and tone control

MarBra:
Add methods to change bass, treble and tone_control.

Verified with AVR-X1600H. Commands sniffed with Denon App on Android.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -333,7 +333,7 @@ class DenonAVR:
         if self._zone == "Main" and add_zones is not None:
             self.create_zones(add_zones)
 
-    def exec_appcommand_post(self, attribute_list, root_tag='tx'):
+    def exec_appcommand_post(self, attribute_list):
         """
         Prepare and execute a HTTP POST call to AppCommand.xml end point.
 
@@ -882,9 +882,7 @@ class DenonAVR:
     def _update_tone_control(self):
         """Update tone control related things."""
         try:
-            root = self.exec_appcommand_post(
-                attribute_list=['GetToneControl'], root_tag='rx'
-                )
+            root = self.exec_appcommand_post(attribute_list=['GetToneControl'])
         except requests.exceptions.ConnectTimeout:
             root = None
 
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -729,7 +729,7 @@ class DenonAVR:
             else:
                 self._state = STATE_OFF
 
-        self._update_tone_control() is False:
+        self._update_tone_control()
 
         return success
 
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -321,6 +321,7 @@ class DenonAVR:
         self._frequency = None
         self._station = None
 
+        self._tone_control_status = None
         self._tone_control_adjust = None
         self._bass = None
         self._bass_level = None
@@ -890,14 +891,20 @@ class DenonAVR:
             _LOGGER.error("Getting tone control failed.")
             return False
 
+        self._tone_control_status = bool(int(root[0].find('status').text))
+        if self._tone_control_status is False:
+            # Tone Control can not be activated.
+            # e.g.: Due to active DynamicEQ
+            return False
+
         try:
             self._tone_control_adjust = bool(int(root[0].find('adjust').text))
             self._bass = int(root[0].find('bassvalue').text)
             self._bass_level = root[0].find('basslevel').text
             self._treble = int(root[0].find('treblevalue').text)
             self._treble_level = root[0].find('treblelevel').text
-        except (AttributeError, IndexError):
-            _LOGGER.error("Incomple/no information found for tone control")
+        except (AttributeError, IndexError, TypeError):
+            _LOGGER.error("Incomplete/no information found for tone control")
             return False
         return True
 
@@ -1940,6 +1947,9 @@ class DenonAVR:
 
     def enable_tone_control(self):
         """Enable tone control to change settings like bass or treble."""
+        if self._tone_control_status is False:
+            return False
+
         if self._tone_control_adjust is True:
             return True
         elif self._set_tone_control_command(parameter_type='adjust', value=1):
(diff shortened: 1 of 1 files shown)


ol-iver:
Thanks 😄 
I'll test tone control later this week with my own receiver

MarBra:
@scarface-4711 did you had some time to look into the PR?

ol-iver:
@MarBra sorry, I forgot about this PR during my vacation. I'll check it on Monday

denonavr/denonavr.py
```diff
@@ -871,6 +879,28 @@ def _get_support_sound_mode(self):
...
+            _LOGGER.error("Getting tone control failed.")
+            return False
+
+        try:
+            self._tone_control_adjust = bool(int(root[0].find('adjust').text))
+            self._bass = int(root[0].find('bassvalue').text)
+            self._bass_level = root[0].find('basslevel').text
+            self._treble = int(root[0].find('treblevalue').text)
+            self._treble_level = root[0].find('treblelevel').text
+        except (AttributeError, IndexError):
```
ol-iver: On my receiver there are two kinds of responses
```xml
<?xml version="1.0" encoding="utf-8" ?>
<rx>
    <cmd>
        <status>0</status>
        <adjust></adjust>
        <basslevel></basslevel>
        <bassvalue></bassvalue>
        <treblelevel></treblelevel>
        <treblevalue></treblevalue>
    </cmd>
</rx>
```
and 
```xml
<?xml version="1.0" encoding="utf-8" ?>
<rx>
    <cmd>
        <status>1</status>
        <adjust>0</adjust>
        <basslevel>0dB</basslevel>
        <bassvalue>6</bassvalue>
        <treblelevel>0dB</treblelevel>
        <treblevalue>6</treblevalue>
    </cmd>
</rx>

The receiver returns the first result, when tone control is not active, like when Dynamic EQ is enabled. This causes a TypeError exception for the type conversions in those lines.

Please change two things:
1. Catch the TypeError exception
2. Check <status> before updating the variables in order to avoid spamming the log with error messages when tone control is not active

```

denonavr/denonavr.py
```diff
@@ -871,6 +879,28 @@ def _get_support_sound_mode(self):
...
+            _LOGGER.error("Incomple/no information found for tone control")
```
ol-iver: "Incomple" is incomplete 😅

denonavr/denonavr.py
```diff
@@ -1865,6 +1915,103 @@ def mute(self, mute):
...
+            return False
+        return True
+
+    def enable_tone_control(self):
+        """Enable tone control to change settings like bass or treble."""
+        if self._tone_control_adjust is True:
+            return True
+        elif self._set_tone_control_command(parameter_type='adjust', value=1):
+            self._tone_control_adjust = True
+            return True
```
ol-iver, line 1966: This check does not cover the case, when tone control cannot be enabled because Dynamic EQ is active.
A way to check this could be to run `self._set_tone_control_command` first, then run self._update_tone_control again and check if `self._tone_control_adjust` is really True

ol-iver (changes requested):
@MarBra sorry it took a while to review your pull request 😅 

I'm able to update tone control on my receiver with your changes. I made to comments, because there are some issues when tone control is not active on the receiver. Could you please have a look at them?

denonavr/denonavr.py
MarBra: > On my receiver there are two kinds of responses
> 
> ```
> <?xml version="1.0" encoding="utf-8" ?>
> <rx>
>     <cmd>
>         <status>0</status>
>         <adjust></adjust>
>         <basslevel></basslevel>
>         <bassvalue></bassvalue>
>         <treblelevel></treblelevel>
>         <treblevalue></treblevalue>
>     </cmd>
> </rx>
> ```
> 
> and
> 
> ```
> <?xml version="1.0" encoding="utf-8" ?>
> <rx>
>     <cmd>
>         <status>1</status>
>         <adjust>0</adjust>
>         <basslevel>0dB</basslevel>
>         <bassvalue>6</bassvalue>
>         <treblelevel>0dB</treblelevel>
>         <treblevalue>6</treblevalue>
>     </cmd>
> </rx>
> 
> The receiver returns the first result, when tone control is not active, like when Dynamic EQ is enabled. This causes a TypeError exception for the type conversions in those lines.
> 
> Please change two things:
> 1. Catch the TypeError exception
> 2. Check <status> before updating the variables in order to avoid spamming the log with error messages when tone control is not active
> ```

I have added `status` as attribute to check, if it is possible to interact with tone control.

denonavr/denonavr.py
MarBra: I have added a check with the `status` attribute to cover the case that tone_control can not be activated.

MarBra:
> @MarBra sorry it took a while to review your pull request 😅
> 
> I'm able to update tone control on my receiver with your changes. I made to comments, because there are some issues when tone control is not active on the receiver. Could you please have a look at them?

@scarface-4711 Thanks for the review. I was able to reproduce the issue also with my receiver. :)

denonavr/denonavr.py
```diff
@@ -871,6 +880,38 @@ def _get_support_sound_mode(self):
...
+            return False
+
+        try:
+            self._tone_control_status = bool(int(root[0].find('status').text))
+        except (AttributeError, IndexError, TypeError):
+            return False
+
+        if self._tone_control_status is False:
+            # Tone Control can not be activated.
+            # e.g.: Due to active DynamicEQ
```
ol-iver, line 901: Wouldn't it make sense to reset the tone control parameters in this case?
```python
self._tone_control_adjust = None
self._bass = None
self._bass_level = None
self._treble = None
self._treble_level = None
```
Activating Dynamic EQ via remote control or web interface deactivates tone control. If the values are not reset, we won't recognize that.

ol-iver (commented):
@MarBra the errors are gone 😃 

Found one last thing. Currently the tone control parameter are not reset, if it is not active. This would make sense, that the library recognizes when someone deactivates it using remote control or the receiver's app.

denonavr/denonavr.py
MarBra: Good catch 👍  Maybe I have sometime in the future to also add Dynamic EQ, so that we can also manage that 😄 
First I will try to bring this tone control feature somehow to Home Assistant 😃

ol-iver (approved):
Looks good, thanks 😃

PR #145, opened 2020-07-05
fix discovery of Denon DN-500AV

starkillerOG:
fixes a HomeAssistant issue where the user is not able to setup the integration:
https://github.com/home-assistant/core/issues/37484


starkillerOG:
@scarface-4711 could you merge this very small PR (only 2 words).
It fixes an issue for a user that is no longer able to use the HomeAssistant integration.

ol-iver:
okay, I'll create a new version including this change

PR #156, opened 2020-09-29
Add auro movie modes

Appesteijn:
On my Denon AVR-X4500H two sound modes are not recognized: AURO-3D and AURO-2D SURROUND

I added those two to Movie and Music mode.


ol-iver:
@Appesteijn thank you 😃 
I will include your changes into the next version

Appesteijn:
👍

ol-iver:
@starkillerOG I'm just starting to wonder if it is intended to add the same sound modes to 'MUSIC' and 'MOVIE'. Wouldn't just the second entry "win" in this case?

starkillerOG:
@scarface-4711 yea don't do it like this, now the mapping form aura to music/movie is undetermined. One of the two will "win" but I am not even sure if the same one will always "win".

Just chose to which Aura schould map to, I would say Dolby Suround maybe? (Instead of movie/music)

Appesteijn:
Shouldn't we then create a new mode AURO to map AURO-3D and AURO-2D SURROUND to? As AURO is a separate 'encoding'? Like this:

https://github.com/scarface-4711/denonavr/commit/7898f7238dba9ac13ea04c1f1356443673fd2653

ol-iver:
According to a [Denon manual](http://manuals.denon.com/AVRX4100W/EU/EN/WBSPSYnbgwnvik.php) it really seems to be an additional sound mode you can buy as an upgrade. According to the manual it is called **Auro-3D**. I would suggest to name it like that.

starkillerOG:
What is important if you can sent an Auro or Auro-3D command to the receiver to actually switch the receiver to that mode.
The status of the receiver is mapped to one of the commands that can be sent to the receiver like MOVIE, MUSIC, STEREO, etc.
Does sending AURO actually switch the receiver to that mode?

Appesteijn:
So with the code:
`('AURO-3D', ['AURO-3D', 'AURO-2D SURROUND']),`
Added to the SOUND_MODE_MAPPING it **doesn't** change to AURO-3D

Please revert my commit until I have figured it out, this is working the way it should. When I added AURO-3D to MUSIC and MOVIE, calling MUSIC and MOVIE showed the sound-mapping-detail menu correctly. But I can't get my receiver to change to AURO-3D directly.

This is what I see in Home Assistant:
```
Logger: DenonAVR
Source: /usr/local/lib/python3.8/site-packages/denonavr/denonavr.py:1716
First occurred: 13:32:19 (2 occurrences)
Last logged: 20:30:15

    Not able to match sound mode: 'AURO-3D', returning raw sound mode.
    Not able to match sound mode: 'AURO-2D SURROUND', returning raw sound mode.
```

Below what I tried from the commandline:
```
~/Dev/denonavr (add_auro_movie_modes) $ python3
Python 3.8.5 (default, Jul 21 2020, 10:48:26) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import denonavr
>>> d = denonavr.DenonAVR("192.168.1.26", "Denon1")
>>> d.update()
True
>>> d.sound_mode_list
['MUSIC', 'MOVIE', 'GAME', 'AUTO', 'STANDARD', 'VIRTUAL', 'MATRIX', 'ROCK ARENA', 'JAZZ CLUB', 'VIDEO GAME', 'MONO MOVIE', 'DIRECT', 'PURE DIRECT', 'DOLBY DIGITAL', 'DTS SURROUND', 'AURO-3D', 'MCH STEREO', 'STEREO', 'ALL ZONE STEREO']
>>> d.set_sound_mode('MUSIC')
True
>>> d.set_sound_mode('MCH STEREO')
True
>>> d.set_sound_mode('AURO-3D')
True
>>> d.set_sound_mode('AURO-2D SURROUND')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mark/Dev/denonavr/denonavr/denonavr.py", line 1661, in set_sound_mode
    self._sound_mode_raw = self._sound_mode_dict[sound_mode][0]
KeyError: 'AURO-2D SURROUND'
>>> d.set_sound_mode('AURO-3D')[0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'bool' object is not subscriptable
>>> d.set_sound_mode('AURO-3D')
True
```

starkillerOG:
@Appesteijn does using `('AURO', ['AURO-3D', 'AURO-2D SURROUND']),` work?

Appesteijn:
Nope. I fixed it in: https://github.com/scarface-4711/denonavr/pull/157

starkillerOG:
@Appesteijn Ah I see, Nice!

PR #163, opened 2020-11-19
Generalize XML commands

JPHutchins:
Here's a followup to @MarBra 's work on the Audyssey commands.  

New schema
For each new XML command class, three constants would be defined in commands.py eg:
```python3
AUDYSSEY_SET_CMD = "SetAudyssey"
AUDYSSEY_GET_CMD = "GetAudyssey"
AUDYSSEY_PARAMS = {
    "dynamiceq": {
        "0": "Off",
        "1": "On",
    },
    "reflevoffset": {
        "0": "0dB",
        "1": "+5dB",
        "2": "+10dB",
        "3": "+15dB",
    },
    "dynamicvol": {
        "0": "Off",
        "1": "Light",
        "2": "Medium",
        "3": "Heavy",
    },
    "multeq": {
        "0": "Off",
        "1": "Flat",
        "2": "L/R Bypass",
        "3": "Reference",
    },
}
```
To define the new command class the init looks like:
```python3
class Audyssey(AppCommand0300):
    """Audyssey Commands."""

    def __init__(self, receiver):
        """
        Initialize Audyssey Settings of DenonAVR.

        :param receiver: DenonAVR Receiver
        :type receiver: DenonAVR
        """

        super().__init__(receiver)
        self._init_commands(
            set_cmd=AUDYSSEY_SET_CMD, get_cmd=AUDYSSEY_GET_CMD, params=AUDYSSEY_PARAMS
        )
        self.dynamiceq = None
        self.multeq = None
        self.reflevoffset = None
        self.dynamicvol = None
```
Note that the attributes defined in AUDYSSEY_PARAMS are initialized here as None - this is their actual state as reported by the receiver and will be updated when the update method is called.

Next, methods would be filled in for each setting, eg:
```python3
def set_multieq(self, setting):
    """Set MultiEQ mode."""
    value = self._param_labels["multeq"].get(setting)
    success = self._set(parameter="multeq", value=value)
    if success:
        self.multeq = setting
```

Finally, the command class must be imported to denonavr.py and initialized:
```python3
from .audyssey import Audyssey, SurroundParameter

self._surround_parameter = SurroundParameter(receiver=self)
```

And expose getters, setters and "listers" to the public interface:
```python3
class DenonAVR:
...
    @property
    def multi_eq(self):
        """Return value of MultiEQ."""
        return self._audyssey.multeq
...
    @property
    def multi_eq_setting_list(self):
        """Return a list of available MultiEQ settings."""
        return self._audyssey.list_parameter_options("multeq")
...
    @multi_eq.setter
    def multi_eq(self, setting):
        """Set MultiEQ."""
        self._audyssey.set_multieq(setting=setting)
```

### DIFF
--- a/denonavr/audyssey.py
+++ b/denonavr/audyssey.py
@@ -11,10 +11,17 @@ import logging
 from io import BytesIO
 import xml.etree.ElementTree as ET
 from requests.exceptions import ConnectTimeout, RequestException
+from denonavr.commands import (
+    COMMAND_ENDPOINT,
+    AUDYSSEY_GET_CMD,
+    AUDYSSEY_PARAMS,
+    AUDYSSEY_SET_CMD,
+    SURROUND_PARAMETER_GET_CMD,
+    SURROUND_PARAMETER_PARAMS,
+    SURROUND_PARAMETER_SET_CMD,
+)
 
 _LOGGER = logging.getLogger("AppCommand0300")
-
-COMMAND_ENDPOINT = "/goform/AppCommand0300.xml"
 
 
 class AppCommand0300:
@@ -28,24 +35,24 @@ class AppCommand0300:
         :type receiver: DenonAVR
         """
         self.receiver = receiver
-        self.param_labels = None
+        self._set_cmd = None
+        self._get_cmd = None
         self._params = None
-        self._get_cmd_name = None
-        self._set_cmd_name = None
+        self._param_labels = None
 
-    def _init_param_labels(self):
-        """Create the map self.param_labels."""
-        if self._params:
-            self.param_labels = {
-                param: {value: key for key, value in inner.items()}
-                for param, inner in self._params.items()
-            }
-        else:
-            _LOGGER.warning("Class AppCommand0300 initialized without params.")
+    def _init_commands(self, set_cmd, get_cmd, params):
+        """Initialize the commands."""
+        self._set_cmd = set_cmd
+        self._get_cmd = get_cmd
+        self._params = params
+        self._param_labels = {
+            param: {value: key for key, value in inner.items()}
+            for param, inner in self._params.items()
+        }
 
     def list_parameter_options(self, param):
         """List the valid options for param."""
-        labels = self.param_labels.get(param)
+        labels = self._param_labels.get(param)
         if labels is None:
             _LOGGER.warning("Parameter: %s not found.", param)
             return []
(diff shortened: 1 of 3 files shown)


denonavr/audyssey.py
```diff
@@ -137,37 +156,93 @@ def _set_audyssey(self, parameter, value):
...
+        self.dynamiceq = None
+        self.multeq = None
+        self.reflevoffset = None
+        self.dynamicvol = None
+
     def dynamiceq_off(self):
         """Turn DynamicEQ off."""
-        if self._set_audyssey(parameter="dynamiceq", value=0) is True:
+        success = self._set(parameter="dynamiceq", value=0)
+        if success:
```
JPHutchins, line 183: Re: all these "if success"
We have a couple options for dealing with state getting out of sync here.  First, it probably makes the most sense to check if the receiver is off - if it is off we can simply return and not update state.  Next, even if it is on and we receive successful reply, we may consider asking for a state update on the modified parameter before setting it.

denonavr/audyssey.py
```diff
@@ -101,29 +111,38 @@ def update(self):
...
-            setattr(
-                self, "{name}_control".format(name=param.get("name")),
-                bool(int(param.get("control"))))
+
+            map_key = self._params.get(name)
+            if not map_key:
+                _LOGGER.warning("Unmapped name: %s", name)
+                continue
+
+            param_text = param.text
```
JPHutchins, line 123: This should handle https://github.com/scarface-4711/denonavr/issues/170

denonavr/audyssey.py
```diff
@@ -101,29 +111,38 @@ def update(self):
...
+            if not param_text:
+                _LOGGER.debug(
+                    "Parameter: %s has no text attribute (state). param control=%s",
+                    param,
+                    param.get("control"),
+                )
+                continue
```
JPHutchins, line 130: I think this handles https://github.com/scarface-4711/denonavr/issues/170

Edit: I am playing a dangerous game where this doesn't work if param.text is an int - reminder to self to update and be more explicit, avoid truthiness problems.

denonavr/denonavr.py
```diff
@@ -339,6 +334,7 @@ def __init__(self, host, name=None, show_all_inputs=False, timeout=2.0,
         self._treble_level = None
 
         self._audyssey = Audyssey(receiver=self)
+        self._surround_parameter = SurroundParameter(receiver=self)
```
JPHutchins, line 337: Initialized here but the update method is never called - it should be added where self._audyssey updates

ol-iver:
I did not forget about this. Last weeks of the year are quite busy, so it will be more a topic for me at Christmas time 😅

JPHutchins:
No rush - this is more for discussion than anything!

ol-iver:
Probably the new async version is covering many of your generalization stuff. I definitely got some inspiration from you. Thanks 😄 
There are too many conflicts to merge this PR with a reasonable effort. Anyway, I happy about your help 👍

PR #168, opened 2020-11-24
Typo fix set_mutlieq in audyssey.py

roblav96:
`set_mutlieq` -> `set_multieq`


JPHutchins:
Good catch - it will need to be corrected where it is called in denon.py, here:
https://github.com/scarface-4711/denonavr/blob/053727bf1ccc09fe629ddf0ea64d84898615167e/denonavr/denonavr.py#L1728-L1731

ol-iver:
Thank you, I included the complete fix in this commit https://github.com/scarface-4711/denonavr/commit/b073203039f1bf2a4b45cd48fd4e9ce298f1f34d

PR #171, opened 2020-12-04
Check for param.text before modifying it

JPHutchins:
Fix for https://github.com/scarface-4711/denonavr/issues/170

If param.text for dynamiceq is None then set self.dynamiceq to None (not False).  This is consistent with what will happen with the other attributes when their respective param.text is None:  `MULTI_EQ_MAP.get(param.text)`

### DIFF
--- a/denonavr/audyssey.py
+++ b/denonavr/audyssey.py
@@ -106,7 +106,7 @@ class Audyssey:
             if param.get("name") == "multeq":
                 self.multeq = MULTI_EQ_MAP.get(param.text)
             elif param.get("name") == "dynamiceq":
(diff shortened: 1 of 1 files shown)


ol-iver:
@JPHutchins thanks 😄 
Changed it slighthly different in this commit https://github.com/scarface-4711/denonavr/commit/b073203039f1bf2a4b45cd48fd4e9ce298f1f34d
In you fix the `self.dynamiceq` would be also None if the receiver would return an integer 0

Additionally I checked for a no None value a couple of lines later https://github.com/scarface-4711/denonavr/blob/master/denonavr/audyssey.py#L119

PR #174, opened 2020-12-16
Fix connectivity issues

JPHutchins:
See https://github.com/home-assistant/core/issues/43670 and elsewhere.

I was wrong about race conditions.  However, if `_get_receiver_sources` fails it was mutating the `self._receiver_type` which caused lots of other requests to fail, timeout 403 etc.

Still not sure why receivers were actually "locking up" - could not recreate even sending batches of 2000 updates with 50 in parallel.  However, during big requests I was able to observe the device info endpoint request time rising from 8ms to 3.3s

We may want to revisit `ensure_configuration` since it still does have a lot of blocking IO that HA doesn't like.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -1084,7 +1084,7 @@ class DenonAVR:
                 root = self.get_status_xml(self._urls.deviceinfo,
                                            suppress_errors=True)
             except (ValueError, requests.exceptions.RequestException):
-                self._receiver_type = None
+                continue
             else:
                 # First test by CommApiVers
                 try:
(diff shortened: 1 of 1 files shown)


ol-iver:
@JPHutchins this change as probably only little effect, because `_get_receiver_sources` is only called when the current input function is not found in `_input_func_list` or when it is not existing at all.

I don't understand your point regarding `ensure_configuration`. It is just performing HTTP requests when there are missing configuration parameter. This is usually only the case when an instance is created. 
Additionally it is only called from `update` method which is exactly the place HA is expecting blocking calls.

Anyhow your change sounds reasonable 😄

PR #180, opened 2021-01-05
add sound mode 'DTS-HD + DOLBY SURROUND'

starkillerOG:


ol-iver:
@starkillerOG merging with the new async version would be too complicated. I added it there. Thx 😄

PR #181, opened 2021-01-05
Add treble_up, treble_down, bass_up, bass_down

MarBra:


MarBra:
@scarface-4711 Any chance to get an review? :) I like to add tonecontrol things to homeassistant and I think down and up methods would make it easier.

ol-iver:
@MarBra could you please include it into the async version? There are too many changes to merge it with a reasonable effort

In the async version there should be an update call after each change. Do you know whether home assistant is calling `update` after each command? I think so, but I'm not 100% sure.
Especially when changing tone control settings, receiver responds with HTTP 200 but did not change anything. So it is better to verify the change with an update instead of assuming it was successfull.

Issue #166, opened 2020-11-21
Media_player entity not available after 0.118 upgrade


mniquette:
I have a denon s730H that has been working fine through 0.117.6. When I upgraded to 0.118  I started the following errors in the log file and the media-player has become unavailable. This has also been reported by several other people on the HA message board in the past day. 

Also the DenonAVR documentation page link to "edit the github" at the top of the page appears to be broken,.

Errors from the log:
Missing status information from XML of Main for: Power, InputFuncSelect, Mute, MasterVolume
3:35:58 PM – /usr/local/lib/python3.8/site-packages/denonavr/denonavr.py (ERROR) - message first occurred at 9:42:07 AM and shows up 2122 times
Host 192.168.1.34 returned HTTP status code 403 to GET request at end point /goform/formMainZone_MainZoneXml.xml
3:35:58 PM – /usr/local/lib/python3.8/site-packages/denonavr/denonavr.py (ERROR) - message first occurred at 9:42:05 AM and shows up 6369 times
Getting renamed and deleted sources failed.
3:35:58 PM – /usr/local/lib/python3.8/site-packages/denonavr/denonavr.py (ERROR) - message first occurred at 9:42:05 AM and shows up 2122 times
Host 192.168.1.34 returned HTTP status code 403 to POST command at end point /goform/AppCommand.xml
3:35:58 PM – /usr/local/lib/python3.8/site-packages/denonavr/denonavr.py (ERROR) - message first occurred at 9:42:05 AM and shows up 2122 times

ol-iver:
There was no change of this library in HA 0.118. 
Did you update the Firmware of your receiver? 
Otherwise please restart it. This helps in many cases.

ScottJWhite:
Well i suppose im not the only one, I've been getting a 403 error for the past week or so. Power cycling the receiver does fix the issue but after a day or 2 the error comes back.

mniquette:
No i haven't updated the firmware. When was the last update to the integration? I will try restarting the unit and see where that takes things.

JPHutchins:
I was thinking misclassification of the receivers causing them to point to the wrong endpoints.  But it's hard to understand when both the /goform/AppCommand.xml and /goform/formMainZone_MainZoneXml.xml are getting 403. 

Coincidentally, I am working on a project to provide a base class for AV receivers that gets "capabilities" from the receivers themselves instead of relying on the complex model number classification.

ol-iver:
[...]

PR #191, opened 2021-02-05
Fix for newer models without DeviceInfo at port 80

elad-bar:
Following thread in HA:
https://github.com/home-assistant/core/issues/43670

After few days the receiver is working DeviceInfo XML is not available for newer models (2016),
Checked it and saw that only after restarting (plug out and in) the receiver the file is getting back to where the DenonAVR is looking for - port 80,
It doesn't try in that case port 8080 which is for new version.

I reversed the order of the types array, if there is a timeout it will not kill the process (as happens today), and only if both types returned timeout it will kill it.

I checked my fix with 2 receivers AVR-X2700 (2016) and AVR-X5200 which is bit older, works perfect.

thanks

### DIFF
--- a/denonavr/foundation.py
+++ b/denonavr/foundation.py
@@ -118,7 +118,9 @@ class DenonAVRDeviceInfo:
         """Identify receiver asynchronously."""
         # Test Deviceinfo.xml if receiver is a AVR-X with port 80 for pre 2016
         # devices and port 8080 devices 2016 and later
-        r_types = [AVR_X, AVR_X_2016]
+        # 2016 models has also some of the XML but not all, try first 2016
+        r_types = [AVR_X_2016, AVR_X]
+
         for r_type in r_types:
             self.api.port = r_type.port
             # This XML is needed to get the sources of the receiver
@@ -129,7 +131,11 @@ class DenonAVRDeviceInfo:
             except AvrTimoutError as err:
                 _LOGGER.debug(
                     "Timeout when identifying receiver", exc_info=err)
-                raise
+
+                # Raise error only when ran over all models
+                if r_types.index(r_type) == len(r_types) - 1:
+                    raise
+
             except AvrRequestError as err:
                 _LOGGER.debug(
                     "Request error when identifying receiver", exc_info=err)


JPHutchins:
Thanks for looking into this!  I will test locally - I actually was not aware that I could get DeviceInfo on port 80 at all, I thought it was only on port 8080.

elad-bar:
@scarface-4711 is there a chance you will review that PR, 
this issue seems to be related also for issues #166 & #169 

thanks

chrismdann:
Is this moving forward?

elad-bar:
@scarface-4711, can you please approve the PR so we will have it in next HA version?

itsamaik:
Same goes for my Marantz SR5013

artemf:
Can confirm the usefulness of the solution.
After powering off/powering on the Marantz SR6012 (2017) and rebooting HA, I've got this in log:
```
Endpoint http://xx.xx.xx.xx:80/goform/Deviceinfo.xml responded with a 403 error. Please consider power cycling your receiver
```
So I do
```
$ curl http://xx.xx.xx.xx:80/goform/Deviceinfo.xml 
Error 403: Forbidden
```
Well, it's expected.

Next I do:
```
$ curl http://xx.xx.xx.xx:8080/goform/Deviceinfo.xml 
<?xml version="1.0" encoding="utf-8" ?>
<Device_Info>
...
```
And it returns the full info.

I does definitely make sense to try both ports, as I have a live example that request on 80 port fails for no reason right after powering off/powering on the AVR.

ol-iver:
I'll definitely check this. There is one obstacle to bring this into HA. The async version uses attrs 20.3.0, but this is pinned to 19.3.0 in HA atm.
We just moved to a new appartment, so my work on this is a bit stuck. I hope this will change soon 😄

elad-bar:
up until that fix will be merged, added mount to the file in docker:
`/homeassistant/py/denonavr/foundation.py:/usr/local/lib/python3.8/site-packages/denonavr/foundation.py:ro`

@scarface-4711 , congrats on the new appartment

ol-iver:
I merged it, but changed the order back to AVR_X and then AVR_X_2016
On a 403 error there will be an `AvrRequestError` exception and the next type will be tested. This is much faster then waiting for a timeout on AVR_X devices in the reversed order

ol-iver:
@elad-bar regarding Denon AVR receivers later than 2016, is the complete API  including the AppCommand.xml endpoint available on 80 and 8080 first or is it just the Deviceinfo.xml?

ol-iver:
btw. I added a functionality which runs setup again once when a 403 error appears for the first time: https://github.com/scarface-4711/denonavr/commit/ba748a7dbb2bc316a283c980105d2d126f3756d6
In case the receiver is shutting down port 80, it should find port 8080 now instead.

PR #193, opened 2021-02-13
Add treble_up, treble_down, bass_up, bass_down

MarBra:
I have added treble_up, treble_down, bass_up, bass_down and a update_tonecontrol method.

### DIFF
--- a/denonavr/tonecontrol.py
+++ b/denonavr/tonecontrol.py
@@ -163,7 +163,7 @@ class DenonAVRToneControl(DenonAVRFoundation):
         if self.bass == 12:
             return
         await self.async_enable_tone_control()
-        await self.async_set_tone_control_command("bassvalue", self.bass + 1 )
+        await self.async_set_tone_control_command("bassvalue", self.bass + 1)
 
     async def async_bass_down(self) -> None:
         """
@@ -175,7 +175,7 @@ class DenonAVRToneControl(DenonAVRFoundation):
         if self.bass == 0:
             return
         await self.async_enable_tone_control()
-        await self.async_set_tone_control_command("bassvalue", self.bass - 1 )
+        await self.async_set_tone_control_command("bassvalue", self.bass - 1)
 
     async def async_set_treble(self, value: int) -> None:
         """
@@ -201,7 +201,9 @@ class DenonAVRToneControl(DenonAVRFoundation):
         if self.treble == 12:
             return
         await self.async_enable_tone_control()
-        await self.async_set_tone_control_command("treblevalue", self.treble + 1 )
+        await self.async_set_tone_control_command(
+            "treblevalue", self.treble + 1
+            )
 
     async def async_treble_down(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


denonavr/tonecontrol.py
```diff
@@ -153,6 +153,30 @@ async def async_set_bass(self, value: int) -> None:
...
+        Increase level of Bass.
+
+        Note:
+        Doesn't work, if Dynamic Equalizer is active.
+        """
+        if self.bass == 12:
+            return
+        await self.async_enable_tone_control()
+        await self.async_set_tone_control_command("bassvalue", self.bass + 1)
+
```
ol-iver, line 168: How about an `await self.async_update()` at this place?
Otherwise the command will work only once.

denonavr/tonecontrol.py
```diff
@@ -153,6 +153,30 @@ async def async_set_bass(self, value: int) -> None:
...
+        Decrease level of Bass.
+
+        Note:
+        Doesn't work, if Dynamic Equalizer is active.
+        """
+        if self.bass == 0:
+            return
+        await self.async_enable_tone_control()
+        await self.async_set_tone_control_command("bassvalue", self.bass - 1)
+
```
ol-iver, line 181: same here with `await self.async_update()`

denonavr/tonecontrol.py
```diff
@@ -167,6 +191,34 @@ async def async_set_treble(self, value: int) -> None:
...
+        Note:
+        Doesn't work, if Dynamic Equalizer is active.
+        """
+        if self.treble == 12:
+            return
+        await self.async_enable_tone_control()
+        await self.async_set_tone_control_command(
+            "treblevalue", self.treble + 1
+            )
+
```
ol-iver, line 210: same here with `await self.async_update()`

denonavr/tonecontrol.py
```diff
@@ -167,6 +191,34 @@ async def async_set_treble(self, value: int) -> None:
...
+        Note:
+        Doesn't work, if Dynamic Equalizer is active.
+        """
+        if self.treble == 0:
+            return
+        await self.async_enable_tone_control()
+        await self.async_set_tone_control_command(
+            "treblevalue", self.treble - 1
+            )
+
```
ol-iver, line 225: same here with `await self.async_update()`

ol-iver (changes requested):
Thanks @MarBra 
There is just a little remark about running updates after up/down commands.

MarBra:
> Thanks @MarBra
> There is just a little remark about running updates after up/down commands.

Thanks for the review @scarface-4711 
I have added the update call to the generic set command. Is this OK for you or shall I add it to the methods itself?

ol-iver:
> 
> 
> > Thanks @MarBra
> > There is just a little remark about running updates after up/down commands.
> 
> Thanks for the review @scarface-4711
> I have added the update call to the generic set command. Is this OK for you or shall I add it to the methods itself?

@MarBra I would prefer it just in the up/down methods themselves, that the update method is just called when it is absolutely neccessary. This follows the way the other interfaces are designed.

MarBra:
> > > Thanks @MarBra
> > > There is just a little remark about running updates after up/down commands.
> > 
> > 
> > Thanks for the review @scarface-4711
> > I have added the update call to the generic set command. Is this OK for you or shall I add it to the methods itself?
> 
> @MarBra I would prefer it just in the up/down methods themselves, that the update method is just called when it is absolutely neccessary. This follows the way the other interfaces are designed.

Alright 👍 Just pushed a new commit.

ol-iver (approved):

ol-iver:
thank you 😄

PR #194, opened 2021-02-13
Add optional attribute to update audyssey settings in main update loop

MarBra:


ol-iver:
Hmmm, I don't see a big benefit in this. The same could be reached by
```python
await denon.async_update()
await denon.async_update_audyssey()
```
Maybe a more generic version like `include_appcommand0300_update` would make sense, when there are multiple classes using that interface. This would enable caching.

MarBra:
> Hmmm, I don't see a big benefit in this. The same could be reached by
> 
> ```python
> await denon.async_update()
> await denon.async_update_audyssey()
> ```
> 
> Maybe a more generic version like `include_appcommand0300_update` would make sense, when there are multiple classes using that interface. This would enable caching.

Yes this is obsolete as we will use an own service for audyssey updates in home-assistant.

PR #212, opened 2021-06-05
reduce log level to warning for Friendly Name not found

starkillerOG:
see: https://github.com/home-assistant/core/issues/51514


ol-iver:
thanks @starkillerOG

PR #213, opened 2021-06-05
add sound mode "DOLBY PL2 CINEMA"

starkillerOG:
see https://github.com/home-assistant/core/issues/51515


ol-iver:
thanks again @starkillerOG 😄

Issue #219, opened 2021-07-12
Questions for updating the docs for async


prcutler:
Hi,

I'm working on updating the README to include information about the new `async` commands (you can see a [WIP in my fork here](https://github.com/prcutler/denonavr/tree/readme-updates))

To avoid the coroutine was never awaited error, to turn the power off and on in the REPL looks like this:

```
>>> await d.async_power_on()
>>> d.async_power_on
>>> await d.async_update()
>>> d.async_update
>>> d.power
'ON'

>>> await d.async_power_off()
>>> d.async_power_off
>>> await d.async_update()
>>> d.async_update
>>> d.power
'OFF'

```

After I run d.async_power_on (or off) or d.async_update the REPL does print out everything I'd ever want to know about my AVR (this is just part of it).  Is this normal?  Are the above commands the right thing to run from the REPL?

Thanks for the confirmation - once I have that, I'll keep working on updating the docs so the usage examples work with `async`.

Paul

```
>>> d.async_update
[...]

PR #220, opened 2021-07-15
README updates for async

prcutler:
Continuing the conversation from Issue #219 here is an updated README.md that includes some of the newer `async` commands.

I couldn't figure out how to add Zone 2 from the asyncio REPL and the async volume commands didn't work for me either in the REPL.  I'd like to update the docs at some point to include those the async way.

It's a start -  this might be helpful for some users that want to know the new async methods. I'm still trying to learn async myself, so there were things I couldn't test.  If anyone wants to give me any pointers to test zones or volume, I'm happy to keep updating the docs.

I broke out the async methods into Power & Input, Sound, and Other methods.  Like the original readme I kept the note to that async methods are preferred and added text that the legacy methods it might be deprecated in the future.

Let me know if you have any feedback or want anything changed.

Thanks!

--P


ol-iver:
thank you 😄

PR #222, opened 2021-08-04
fix friendly name warning

starkillerOG:
some AVRs do not have the fiendly name tag, specifically the Denon AVR 2312, however it is present in the device_info xml.
So invert the order of requesting to silence the friendly name warning since it is then already retrieved
see https://github.com/home-assistant/core/issues/51514


starkillerOG:
@scarface-4711 could you merge this?

PR #223, opened 2021-08-05
Fixed All Zone Stereo nog getting turned off

Tommatheussen:
I was having this weird issue where ALL ZONE STEREO was not being turned off, unless I power cycled the received. Manually calling the `MNZST OFF` command via the Home Assistant service call does work and correctly turn off the all zone stereo.

**NOTE**: Haven't actually tested this change just yet
**EDIT**: Just tested and confirmed it works


ol-iver:
You're right. That's a bug. Thanks for fixing 😄

PR #224, opened 2021-08-10
Add DTS-HD + DSUR as a sound type

lewisporter:
I've been getting this warning in home assistant when using the integration with my Denon AVC-X3700H:
`Not able to match sound mode: 'DTS-HD + DSUR', returning raw sound mode.`

From looking at #106 , it seems that the issue is a simple matter of a missing enum. This ought to fix that.


ol-iver:
thanks 👍

PR #235, opened 2022-06-20
README: link denonavr-cli as a client

mgorny:
First of all, thanks for the awesome library! I've just started a new [denonavr-cli](https://pypi.org/project/denonavr_cli/) project that provides a trivial CLI on top of it, so I've figured you might be interested in including it as the second entry in the users list ;-).


ol-iver:
cool 😄

mgorny:
Thanks!

PR #238, opened 2022-10-25
Telnet status

dcmeglio:
Your integration is great, and using HTTP is the right way to go for a number of reasons, but it has one major drawback when used in Home Assistant which is that the only way to get device status updates is to poll the device. By default this means updates can be delayed up to 30s. However, Denon also provides real-time push notifications via TCP on port 23. This PR adds methods to connect on port 23 and to create a task to monitor the TCP stream for real-time updates which are then pushed to the subscriber by callback methods. 

I added an example to the README but essentially:

```
import asyncio
import denonavr
d = denonavr.DenonAVR("192.168.1.119")
await d.async_setup()
await d.async_update()
await d.async_connect()
d.register_callback(_update_callback)
await asyncio.wait([d.monitor_updates()])
def _update_callback(device):
   print(device.volume)
```

This is really my first time working in python so very much interested in feedback and not even sure if you're interested in merging this into this library since everything you have right now is HTTP based, but I figured I'd see so I don't have to maintain my own fork!

Edit: I'm still trying to figure out how you manage the playing state so if you can give any pointers on how I might integrate that into the PW and SI commands I'm all ears! I think I need to call `async_update_media_state` when the PW or SI command executes, but not sure what I should specify for the `cache_id`


JPHutchins:
Hi @dcmeglio !  This same desire for an asynchronous "push" interface led me to create the pyavreceiver project:  https://github.com/JPHutchins/pyavreceiver

When integrated with Home Assistant, for example, you can see the volume slider in Home Assistant update in real time as you physically move the volume knob of the receiver.  Since pyavreceiver implements most (all?) of the Denon telnet API, the same would apply to most parameters of the receiver.

I'd highly value your testing and contribution to the project if you are interested.

Cheers,
J.P.

dcmeglio:
Nice! Is there an HA component already?

JPHutchins:
Yeah, it was in progress 1.5 years ago.  I'm sure stuff with HA has changed since then so it might be easier to start fresh than to rebase, IDK: https://github.com/home-assistant/core/pull/45071

dcmeglio:
@ol-iver any thoughts on this? I know it's a bigger change so not a super quick PR review but I'd love to be able to get this up on pypi so I can submit a PR against the HA integration too. Any feedback? Like I said, first time I've built some async code in pything before

ol-iver:
@dcmeglio thanks for creating this PR 😄 
We could go ahead with your telnet approach. It could be an optional way of updating the properties in HA.
However, you put all the things into `DenonAVR` class which should not the way to go. Thus, please apply several changes before we can merge things:
1. Create a new class like `DenonAVRTelnetAPI` in `api.py` and move all telnet connection related stuff and the abilty to register callbacks here. It might make sense to register callbacks event specific.
2. Add this class to `DenonAVRDeviceInfo` similiar to how it is done for the [REST API](https://github.com/ol-iver/denonavr/blob/38bb1641202411cafd34cfc751de835fa403e77f/denonavr/foundation.py#L41-L44)
3. Move all code which is changing properties like volume, soundmode etc. into callbacks of the respective classes and register them at `DenonAVRTelnetAPI`.
4. `async_connect` should immediately start watching for changes and should remain non-blocking. Please create a `async_disconnect` method to stop watching. This might be done by calling `monitor_updates` from `async_connect` and saving the resulting future in order to  `cancel()` it in `async_disconnect` .
5. Could you provide test cases?

Please use `attrs` for any new class and attribute as it is done in the rest of the library.
`cache_id` is not important in the telnet case. It is set in the regular `async_update` case, that API calls can be cached, when different functions are calling the same endpoint.

dcmeglio:
Yeah I'll give it a go. Might require a few back and forths, like I said I'm still learning Python

dcmeglio:
A couple questions as I work through this 

> Create a new class like DenonAVRTelnetAPI in api.py and move all telnet connection related stuff and the abilty to register callbacks here

I'm moving the callback registrations here, but was your thought that HA also calls this directly to register a callback? Or that DenonAVR would also expose a register_callback type method? To get alerted of a pushed update HA needs to register to receive the callback

> Move all code which is changing properties like volume, soundmode etc. into callbacks of the respective classes and register them at DenonAVRTelnetAPI.

Not totally sure I understand this one. You want soundmode, for example, to register for a callback on the api and then use that to update itself?

Edit: If it's helpful to visualize I'm thinking something like this in the DenonAVRVolume class:

    def setup(self) -> None:
        """Ensure that the instance is initialized."""
        # Add tags for a potential AppCommand.xml update
        for tag in self.appcommand_attrs:
            self._device.api.add_appcommand_update_tag(tag)

        self._device.telnet_api.register_volume_callback(self._volume_callback)
        self._is_setup = True

    def _volume_callback(self, value: float) -> None:
        self._volume = value

ol-iver:
Just as you pointed it out in your example. Maybe a generic `register_callback` method instead of `register_volume_callback`. Registration could be event specific like `self._device.telnet_api.register_callback("MV", self._volume_callback)`
I don't think HA needs to register any callback, so let's start with a registration for library internal callbacks only,

dcmeglio:
I think HA does need to register an "overall" callback so it knows "hey the library received an update from the device, refresh the HA state" it doesn't necessarily need to know it was an MV or an MU, it just needs to know **something** changed so it writes the new state to the state machine

ol-iver:
Let's check in existing HA components which are using push, how they are notifying HA when something changed. I never did that before..

dcmeglio:
Me either, let me get through this piece first but I had asked on the HA discord and was given this as an example https://github.com/oyvindwe/home-assistant/blob/dev/homeassistant/components/nobo_hub/sensor.py#L77 https://github.com/oyvindwe/home-assistant/blob/dev/homeassistant/components/nobo_hub/sensor.py#L92-L95

Making good progress on the refactor. Should have most of this done in the next few days. For some unit tests, I see you have some examples I can work off of but stupid question as a new python guy... how do I run your existing tests?

dcmeglio:
Items 1-4 are addressed. I made a super basic test at this point. Definitely could use more but I'm still trying to figure out the testing framework. Figured I'd at least submit the rest to get your opinion on the refactor

dcmeglio:
Any update on the pr review?

ol-iver:
I'll review this weekend 😄

dcmeglio:
> I'll review this weekend 😄

Any chance you were able to take a look? :)

.gitignore
```diff
@@ -98,3 +98,5 @@ ENV/
 
 # Visual Studio Code
 .vscode/
+
+sample.py
```
ol-iver: Why that?

README.md
```diff
@@ -50,6 +54,20 @@ The `asyncio` library should automatically be imported in the REPL.  Import the
...
+>>> import asyncio
+>>> import denonavr
+>>> d = denonavr.DenonAVR("192.168.1.119")
+>>> await d.async_setup()
+>>> await d.async_update()
+>>> await d.async_connect()
+>>> d.register_callback(_update_callback)
+>>> await asyncio.wait([d.monitor_updates()])
+>>> def _update_callback(device):
+>>>>>> print(device.volume)
```
ol-iver: `_update_callback` needs to be defined before registering it in line 65.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+
+    async def _trigger_callbacks(self, type: str, zone: str, value: any):
+        """Handle triggering the registered callbacks for the specified type"""
+        if type in self._callbacks.keys():
+            for callback in self._callbacks[type]:
+                try:
+                    if inspect.iscoroutinefunction(callback):
+                        await callback(zone, value)
+                    else:
+                        callback(zone, value)
```
ol-iver: Please don't `inspect` on each callback call, but do it when registering the callbacks. 
I would prefer to allow coroutines only as callbacks, because `_trigger_callbacks` is asynchronous too. Otherwise, there is always danger that some adds a function which blocks the event loop.

Additionally, please call the callbacks like this ```await callback(event, parameter)```. We could share the same interface for all callbacks then.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+            self._callbacks[type] = []
+        self._callbacks[type].append(callback)
+
+    def unregister_callback(self, type: str="ALL", callback=lambda *args: any):
+        """Unregisters a callback handler for an event type."""
+        if not type in self._callbacks.keys():
+            return
+        self._callbacks[type].remove(callback)
+
+    async def _process_event(self, message):
```
ol-iver: type annotation of `message` is missing.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+                        data += chunk[i].to_bytes(1, byteorder='big')
+                    else:
+                        await self._process_event(str(data,'utf-8'))
+                        data = bytearray()
+            except asyncio.exceptions.TimeoutError as err:
+                _LOGGER.debug("Lost connection to receiver, reconnecting")
+                await self.async_disconnect()
+                await self.async_connect()
+
+    def register_callback(self, type: str="ALL", callback=lambda *args: any):
```
ol-iver: Why adding defaults here? The annotation does not look right too.
```suggestion
    def register_callback(self, type: str, callback: Callable[[str, str], Awaitable[None]]):
```
Please allow registering Coroutines only. See my comment at `_trigger_callbacks`.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+        """Registers a callback handler for an event type."""       
+        if not type in self._callbacks.keys():
+            self._callbacks[type] = []
+        self._callbacks[type].append(callback)
```
ol-iver: Could you check the types? There are only a bunch of types which are working.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+
+    def unregister_callback(self, type: str="ALL", callback=lambda *args: any):
```
ol-iver:
```suggestion
    def unregister_callback(self, type: str, callback: Callable[[str, str], Awaitable[None]]):
```

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+            else:
+                zone = ZONE3
+            if parameter == 'ON' or parameter == 'OFF':
+                await self._process_power(zone, parameter)
+            elif parameter == 'MUON' or parameter == 'MUOFF':
+                await self._process_mute(zone, parameter)
+            elif parameter in ZONE_SOURCES:
+                await self._process_input(zone, parameter)
+            elif parameter.isdigit():
+                await self._process_volume(zone, parameter)
```
ol-iver: This is way too much logic for an API class. 
Please just run something like `await self._run_callbacks(event, parameter)` and everything else in the classes which are handling those properties.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+            if parameter == 'ON' or parameter == 'OFF':
+                await self._process_power(zone, parameter)
+            elif parameter == 'MUON' or parameter == 'MUOFF':
+                await self._process_mute(zone, parameter)
+            elif parameter in ZONE_SOURCES:
+                await self._process_input(zone, parameter)
+            elif parameter.isdigit():
+                await self._process_volume(zone, parameter)        
+
+    async def _trigger_callbacks(self, type: str, zone: str, value: any):
```
ol-iver:
```suggestion
    async def _run_callbacks(self, event: str, parameter: str):
```
Please see previous comment

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+                    # We don't want a single bad callback to trip up the whole system and prevent further 
+                    # execution
+                    print(err)
+                    _LOGGER.error(f"Event callback triggered an unhandled exception {err}")
+
+        if "ALL" in self._callbacks.keys():
+            for callback in self._callbacks["ALL"]:
+                try:
+                    if inspect.iscoroutinefunction(callback):
+                        await callback(zone)
```
ol-iver:
```suggestion
                        await callback(event, parameter)
```
The same in regards of `inspect` as in previous comment.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+            value = parameter[7:]
+            await self._trigger_callbacks("MULTEQ", zone, value)
+        elif parameter == "DYNEQ ON":
+            await self._trigger_callbacks("DYNEQ", zone, "1")
+        elif parameter == "DYNEQ OFF":
+            await self._trigger_callbacks("DYNEQ", zone, "0")
+        if parameter == "TONE CTRL OFF":
+            await self._trigger_callbacks("TONE", zone, "1")
+        elif parameter == "TONE CTRL ON":
+            await self._trigger_callbacks("TONE", zone, "0")
```
ol-iver: All these functions belong into the classes which are handling the properties.

denonavr/audyssey.py
```diff
@@ -58,6 +58,13 @@ class DenonAVRAudyssey(DenonAVRFoundation):
     appcommand0300_attrs = {
         AppCommands.GetAudyssey: None}
 
+    def __attrs_post_init__(self) -> None:
+        """Initialize the callbacks"""
+        self._device.telnet_api.register_callback("REFLEV", self._reflevoffset_callback)
+        self._device.telnet_api.register_callback("DYNVOL", self._dynamicvol_callback)
+        self._device.telnet_api.register_callback("MULTEQ", self._multeq_callback)
+        self._device.telnet_api.register_callback("DYNEQ", self._dynamiceq_callback)
```
ol-iver: Please register callbacks in `setup` method instead.

denonavr/foundation.py
```diff
@@ -82,6 +87,13 @@ def __attrs_post_init__(self) -> None:
         else:
             raise ValueError("Invalid zone {}".format(self.zone))
 
+        self.telnet_api.register_callback("PW", self._power_callback)
```
ol-iver: Please register callbacks in `async_setup` method instead.

denonavr/foundation.py
```diff
@@ -112,6 +124,8 @@ async def async_setup(self) -> None:
             self.api.add_appcommand_update_tag(
                 AppCommands.GetAllZonePowerStatus)
 
+            await self.telnet_api.async_connect()
```
ol-iver: Telnet should be optional. You don't know if every receiver supports telnet. Additionally, telnet supports only one connection per receiver and there might be cases where multiple instances of denonavr are running in parallel.

I would prefer if we just have two functions like `async_telnet_connect` and `async_telnet_disconnect` here.

denonavr/volume.py
```diff
@@ -55,6 +55,11 @@ class DenonAVRVolume(DenonAVRFoundation):
         "_volume": "./MasterVolume/value",
         "_muted": "./Mute/value"}
 
+    def __attrs_post_init__(self) -> None:
+        """Initialize the callbacks"""
+        self._device.telnet_api.register_callback("MV", self._volume_callback)
+        self._device.telnet_api.register_callback("MU", self._mute_callback)
```
ol-iver: Please register callbacks in `setup` method instead.

denonavr/volume.py
```diff
@@ -24,7 +24,7 @@
 
 def convert_muted(value: str) -> bool:
     """Convert muted to bool."""
-    return bool(value == STATE_ON)
+    return bool(value == STATE_ON or value == STATE_ON_CAPS)
```
ol-iver:
```suggestion
    return bool(value.lower() == STATE_ON)
```

denonavr/const.py
```diff
@@ -252,6 +252,8 @@
 POWER_ON = "ON"
 POWER_OFF = "OFF"
 POWER_STANDBY = "STANDBY"
+STATE_ON_CAPS = "ON"
+STATE_OFF_CAPS = "OFF"
```
ol-iver: We don't need these constants. Please see my comment in `volume.py`

denonavr/input.py
```diff
@@ -140,6 +140,11 @@ class DenonAVRInput(DenonAVRFoundation):
     status_xml_attrs = {
         "_input_func": "./InputFuncSelect/value"}
 
+    def __attrs_post_init__(self) -> None:
+        """Initialize the callbacks"""
+        self._device.telnet_api.register_callback("SI", self._input_callback)
+        self._device.telnet_api.register_callback("PW", self._power_callback)
```
ol-iver: Please register callbacks in setup method instead.

denonavr/const.py
```diff
@@ -262,6 +264,12 @@
 ZONE2 = "Zone2"
 ZONE3 = "Zone3"
 VALID_ZONES = [MAIN_ZONE, ZONE2, ZONE3]
+ZONE_SOURCES = ["CD", "PHONO", "TUNER", "DVD", "BD", "TV", 
+    "SAT/CBL","MPLAY","GAME","HDRADIO","NET","PANDORA",
+    "SIRIUSXM","LASTFM","FLICKR","IRADIO","SERVER","FAVORITES",
+    "AUX1","AUX2","AUX3","AUX4","AUX5","AUX6","AUX7","BT","USB/IPOD",
+    "USB DIRECT", "IPOD DIRECT"
+]
```
ol-iver: Please call this constant `TELNET_SOURCES` and put it right behind `CHANGE_INPUT_MAPPING` constant.

denonavr/denonavr.py
```diff
@@ -218,6 +219,12 @@ def get_command(self, request: str) -> str:
     def send_get_command(self, request: str) -> str:
         """Send HTTP GET command to Denon AVR receiver...for compatibility."""
 
+    def register_callback(self, callback: lambda *args: any):
```
ol-iver:
```suggestion
    def register_callback(self, type: str, callback: Callable[[str, str], Awaitable[None]]):
```
Why not giving everybody all the options?

denonavr/denonavr.py
```diff
@@ -218,6 +219,12 @@ def get_command(self, request: str) -> str:
...
+        self._device.telnet_api.register_callback(callback=callback)
+
+    def unregister_callback(self, callback:lambda *args: any):
```
ol-iver:
```suggestion
    def unregister_callback(self, type: str, callback: Callable[[str, str], Awaitable[None]]):
```

README.md
```diff
@@ -50,6 +54,20 @@ The `asyncio` library should automatically be imported in the REPL.  Import the
...
 ```
 
+### Monitoring with TCP
+```
+>>> import asyncio
+>>> import denonavr
+>>> d = denonavr.DenonAVR("192.168.1.119")
+>>> await d.async_setup()
+>>> await d.async_update()
+>>> await d.async_connect()
```
ol-iver: This does not fit to your implementation 😅

ol-iver (commented):
You made very good progress 🚀 
I went almost through it, but did not check the tests yet and did not test it with my own receiver. These are my remarks up to that point.
There are a couple of tests failing to. It might make sense to run them locally. You can find [here](https://github.com/ol-iver/denonavr/blob/main/.github/workflows/python-tests.yml) which tests are running.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+        try:
+            self._socket_reader, self._socket_writer = await asyncio.wait_for(asyncio.open_connection(self.host, 23), timeout=self.timeout)
+            self._telnet_task = asyncio.create_task(self._async_monitor())
+            return self._telnet_task
+        except asyncio.exceptions.TimeoutError as err:
+            _LOGGER.debug(
+                "Socket timeout exception on connect",
+                exc_info=True)
+            raise AvrTimoutError(
+                "TimeoutException: {}".format(err), "connect") from err
```
ol-iver:
```suggestion
        """ Connect to the receiver asynchronously."""
        try:
            self._socket_reader, self._socket_writer = await asyncio.wait_for(asyncio.open_connection(self.host, 23), timeout=self.timeout)
        except asyncio.exceptions.TimeoutError as err:
            _LOGGER.debug(
                "Socket timeout exception on connect",
                exc_info=True)
            raise AvrTimoutError(
                "TimeoutException: {}".format(err), "connect") from err
        self._telnet_task = asyncio.create_task(self._async_monitor())
        return self._telnet_task
```
It looks cleaner if you only put the statements which could raise the exception between `try` and `except`.

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+                    # Messages are CR terminated
+                    if chunk[i] != 13:
+                        data += chunk[i].to_bytes(1, byteorder='big')
+                    else:
+                        await self._process_event(str(data,'utf-8'))
+                        data = bytearray()
+            except asyncio.exceptions.TimeoutError as err:
+                _LOGGER.debug("Lost connection to receiver, reconnecting")
+                await self.async_disconnect()
+                await self.async_connect()
```
ol-iver:
```suggestion
        """Reads the messages on the TCP socket."""
        data = bytearray()
        while not self._socket_reader.at_eof():
            try:
                chunk = await asyncio.wait_for(self._socket_reader.read(_SOCKET_READ_SIZE), self.timeout)
            except asyncio.exceptions.TimeoutError as err:
                _LOGGER.debug("Lost connection to receiver, reconnecting")
                await self.async_disconnect()
                await self.async_connect()
            for i in range(0,len(chunk)):
                # Messages are CR terminated
                if chunk[i] != 13:
                    data += chunk[i].to_bytes(1, byteorder='big')
                else:
                    await self._process_event(str(data,'utf-8'))
                    data = bytearray()
```
Same here

denonavr/api.py
```diff
@@ -309,3 +313,184 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+            for callback in self._callbacks[type]:
+                try:
+                    if inspect.iscoroutinefunction(callback):
+                        await callback(zone, value)
+                    else:
+                        callback(zone, value)
+                except Exception as err:
+                    # We don't want a single bad callback to trip up the whole system and prevent further 
+                    # execution
+                    print(err)
```
ol-iver: Don't `print`, use `_LOGGER` only

ol-iver (commented):
Some things I forgot to mention 😄

denonavr/volume.py
dcmeglio: I originally had it there but that doesn’t seem to work for zones 2 and 3.

ol-iver: Did you call `async_setup` for each zone?

.gitignore
dcmeglio: This was just something I was using for my local testing. I meant to exclude it from the gitignore but clearly forgot

denonavr/api.py
dcmeglio: I can do that, the problem is it seemed like this would be a poor separation of concerns since the individual class (e.g. mute) has to know how to parse logic about the overarching API message. Meaning the mute class now becomes tightly coupled.

denonavr/foundation.py
dcmeglio: Agreed on the optional piece. Can you explain the `I would prefer if we just have two functions like async_telnet_connect and async_telnet_disconnect here.`? Are you saying you'd rather the caller have to call async_telnet_connect themselves?

denonavr/volume.py
dcmeglio: Your logic (I didn't dig in too deeply) seemed to purposely not do that. Perhaps I'm wrong but it seems to be intentional that this is called for only zone 1

denonavr/denonavr.py
dcmeglio: Great point. I was originally thinking specifically of how this would be used in HA which is shortsighted.

denonavr/api.py
dcmeglio: You're suggesting validating the `type` against the list of commands we care about? Yeah I could do that.

denonavr/api.py
dcmeglio: Oops, left over debugging code.

denonavr/api.py
dcmeglio: I can do that if you'd like, not a huge fan of burying the API implementation details in the individual classes but I can.

denonavr/api.py
dcmeglio: I went with a compromise here that I hope you'll agree makes sense. I believe the Z2 vs Z3 handling is a detail of the overall API and should not be handled internally by each command. Meaning, I believe that is an implementation detail of the api not of volume vs mute vs etc. So I think we should leave this in api.py but other than that exception, now all command processing logic is in the individual classes. What do you think?

Edit: Oh I also left the special handling of the MVMAX message because everything I can find online indicates this is junk.  No Denon documentation I can find acknowledges it exists and it seems to do nothing.

dcmeglio:
@ol-iver I believe I resolved all the linter issues causing the tests to fail. I also believe I addressed all your comments. I pretty much made the changes you requested except I think I came up with a better approach regarding the command handling and zones to properly separate the concerns of overall API handling vs. individual command processing.

Think you might be able to give this a re-review? I'd really love to get this merged soon so I have a shot of getting a PR together for HA to support this in the 2023.1 release!

dcmeglio:
Any luck taking a look at this soon? I'd really like to see if we can get this updated in HA. Right now the lack of a real time update is causing my home theater automations to not work properly so I think this could be a nice addition for myself and others.

dcmeglio:
I'm off from work this week so I was hoping if there were any remaining issues to be able to tackle them quickly. Any chance you can re-review this week?

dcmeglio:
Is there any update here? I know you must be busy, but I really want to get this published and I'm not sure if you're going to request further changes. I was really hoping we could get the change in HA 2023.1 but that's not happening now, do you think we can shoot for 2023.2? Once you approve this, I think the changes to the HA component will be pretty minimal

ol-iver:
I was not busy, but I was skiing for the last weeks ⛷️  and there are no computer allowed on the slopes 😉 
I'll check your PR this week 😄

dcmeglio:
Cool! Hope you enjoyed, I skied once and fell hard and never did it again :) Let me know if any extra changes are needed.

dcmeglio:
Any news?

denonavr/api.py
```diff
@@ -309,3 +314,148 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+@attr.s(auto_attribs=True, hash=False, on_setattr=DENON_ATTR_SETATTR)
+class DenonAVRTelnetApi:
+    """Handle Telnet responses from the Denon AVR Telnet interface."""
+
+    host: str = attr.ib(converter=str, default="localhost")
+    timeout: int = attr.ib(converter=int, default=10)
+    _telnet_task: asyncio.Task = attr.ib(default=None)
+    _socket_reader: asyncio.StreamReader = attr.ib(default=None)
+    _socket_writer: asyncio.StreamWriter = attr.ib(default=None)
+    _callbacks: dict[str, Callable] = attr.ib(
```
ol-iver:
```suggestion
    _callbacks: Dict[str, Callable] = attr.ib(
```

denonavr/api.py
```diff
@@ -309,3 +314,148 @@ def prepare_appcommand_body(cmd_list: Tuple[AppCommandCmd]) -> bytes:
...
+            if parameter in ('ON', 'OFF'):
+                await self._run_callbacks("PW", zone, parameter)
+            elif parameter in ('MUON', 'MUOFF'):
+                await self._run_callbacks("MU", zone, parameter)
+            elif parameter in TELNET_SOURCES:
+                await self._run_callbacks("SI", zone, parameter)
+            elif parameter.isdigit():
+                await self._run_callbacks("MV", zone, parameter)
+        elif event in TELNET_EVENTS:
+            await self._run_callbacks(event, MAIN_ZONE, parameter)
```
ol-iver:
```suggestion
    async def _process_event(self, message: str):
        """Process a realtime event."""
        if len(message) < 3:
            return None
        zone = MAIN_ZONE

        # Event is 2 characters
        event = message[0:2]
        # Parameter is the remaining characters
        parameter = message[2:]
        
        if event == "MV":
            # This seems undocumented by Denon and appears to basically be a
            # noop that goes along with volume changes. This is here to prevent
            # duplicate callback calls.
            if parameter[0:3] == "MAX":
                return

        if event in ("Z2", "Z3"):
            if event == "Z2":
                zone = ZONE2
            else:
                zone = ZONE3

            if parameter in ("ON", "OFF"):
                event = "PW"
            elif parameter in TELNET_SOURCES:
                event = "SI"
            elif parameter.isdigit():
                event = "MV"
            elif parameter[0:2] in TELNET_EVENTS:
                event = parameter[0:2]
                parameter = parameter[2:]

        if event not in TELNET_EVENTS:
            return

        await self._run_callbacks(event, zone, parameter)
```
Maybe something like this to have less `self._run_callbacks` calls.
Additionally, for zones most of the events could be just shifted by 2 letters. This would cover some more commands of the zones which seem to exist according to [this document](https://www.heimkinoraum.de/upload/files/product/IP_Protocol_AVR-Xx100.pdf).

denonavr/audyssey.py
```diff
@@ -64,8 +64,26 @@ def setup(self) -> None:
...
+            if parameter[0:6] == "REFLEV":
+                self._reflevoffset = parameter[7:]
+            elif parameter[0:6] == "DYNVOL":
+                self._dynamicvol = parameter[7:]
+            elif parameter[0:6] == "MULTEQ":
+                self._multeq = parameter[7:]
+            elif parameter == "DYNEQ ON":
+                self._dynamiceq = "1"
+            elif parameter == "DYNEQ OFF":
+                self._dynamiceq = "0"
```
ol-iver:
```suggestion
        if self._device.zone != zone:
            return

        if parameter[0:6] == "REFLEV":
            self._reflevoffset = parameter[7:]
        elif parameter[0:6] == "DYNVOL":
            self._dynamicvol = parameter[7:]
        elif parameter[0:6] == "MULTEQ":
            self._multeq = parameter[7:]
        elif parameter == "DYNEQ ON":
            self._dynamiceq = "1"
        elif parameter == "DYNEQ OFF":
            self._dynamiceq = "0"
```
The same for the other callbacks where `zone` is checked.

README.md
```diff
@@ -50,6 +54,20 @@ The `asyncio` library should automatically be imported in the REPL.  Import the
...
+>>> import asyncio
+>>> import denonavr
+>>> d = denonavr.DenonAVR("192.168.1.119")
+>>> await d.async_setup()
+>>> await d.async_telnet_connect()
+>>> await d.async_update()
+>>> def _update_callback(zone, event, parameter):
+>>>>>> print(d.volume)
+>>> d.register_callback(_update_callback)
+>>> await asyncio.wait([d.monitor_updates()])
```
ol-iver:
```suggestion
>>> def update_callback(zone, event, parameter):
>>>>>> print("Zone: " + zone + " Event: " + event + " Parameter: " + parameter)
>>> d.register_callback("ALL", update_callback)
```

ol-iver (commented):
I made it 😅 
This looks good now, thanks 👍 
There are only a couple of small things left.

dcmeglio:
I approved all your suggestions and made the corresponding zone check suggestion in the other files to reduce nesting. Also fixed one small linter issue. I think it has everything you were looking for!

test-requirements.txt
```diff
@@ -6,4 +6,5 @@ pytest-timeout
 pytest-asyncio
 pytest-httpx
 flake8-docstrings
-flake8
\ No newline at end of file
+flake8
+asynctest
```
ol-iver: Could you please remove `asynctest`? It hasn't been updated for 3 years and does not work with Python 3.11 anymore.

tests/test_denonavr.py
```diff
@@ -142,74 +146,237 @@ def custom_matcher(self, request: httpx.Request, *args, **kwargs):
...
+            debug_mock.return_value = (reader, asyncio.StreamReader())
+            httpx_mock.add_callback(self.custom_matcher)
+
+            self.denon = denonavr.DenonAVR(FAKE_IP)
+            self.future = asyncio.Future()
+            self.denon.register_callback("ALL", self._callback)
+            await self.denon.async_setup()
+            await self.denon.async_telnet_connect()
+            reader.feed_data(b"PWON\r")
+            await asyncio.sleep(0.5)
```
ol-iver: Why sleeping here, instead of waiting for the callback as in the other tests?

ol-iver (commented):
There is one more thing, I found when testing different python versions.
`asynctest` hasn't been updated for 3 years and it does not work in python 3.11 anymore. I don't think that it is a good idea to introduce it now. Could you patch `open_connection` in a different way?

test-requirements.txt
dcmeglio: Agreed, I hadn't done tests like this in python before so I was just following some suggestions from google. Any chance you know of another library? Otherwise I'll find one.

tests/test_denonavr.py
dcmeglio: This was a mistake. I had done it for all the tests this way before I learned about waiting on the future. I missed rewriting this one. I corrected this test. I'll see what other options there are instead of asynctest but if you have suggestions, unit testing in python is new to me.

test-requirements.txt
ol-iver: It might work simply replacing `asynctest.patch` by `unittest.mock.patch`

dcmeglio: Couple other small tweaks needed to use that, but yup that did it. PR updated.

ol-iver: The tests seem to work, but there is a pylint error now

dcmeglio:
My bad, just did a find/replace and didn't rerun the linter. Please hold.

dcmeglio:
Hopefully that does it 🤞

test-requirements.txt
dcmeglio: Can you run it now? For some reason I can't run pylint locally against the tests directory so I'm unfortunately resolving the issues manually:

```
************* Module tests
tests\__init__.py:1:0: F0010: error while code parsing: Unable to load file tests\__init__.py:
[Errno 2] No such file or directory: 'tests\\__init__.py' (parse-error)
```

dcmeglio:
Now I recall what I read. Here's the problem, unittest.mock in < 3.8 did not support async. That was why I used asynctest. AsyncMock support was added to unitest.mock in 3.8 `Changed in version 3.8: [patch()](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) now returns an [AsyncMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock) if the target is an async function.` 

So not sure how to handle this? Prior to 3.8 it seems `asynctest` was the way to go

dcmeglio:
There might be a way to conditionally load asynctest just for 3.7. Still reading. If you have other ideas though let me know.

ol-iver:
That was annoying stuff 😅 
I just implemented a small class which supports mocking async call in python 3.7. It includes just a few lines.
Additionally, I removed the mocks from tests where telnet is not used. I think we are good to go now 😄 
Thanks for your effort 🚀

dcmeglio:
Awesome! I'm going to start working on the HA changes to use this. Just curious if you have any idea when you might publish a new pypi version?

ol-iver:
I changed the handling of the telnet connection.
- the reconnect was not working, because it canceled its own task
- the timeout when reading from the socket was not correct. It was not a connection timeout, but it terminated the connection when there was no traffic for a while.

@dcmeglio Please double-check with your receiver.

dcmeglio:
Hmm the reconnect was working for me with my code (maybe the way I was simulating the socket error was different?), but regardless everything is working fine with your changes including the reconnect. Thanks for the improvements, helping me learn python sockets a bit better too.

dcmeglio:
Also just tested my proposed Home-Assistant PR with your latest changes, everything is working good there  too 👍

PR #239, opened 2023-01-16
Fix issue where callback methods weren't marked as async

dcmeglio:
Discovered a small bug while working on the HA integration. After removing support for synchronous callbacks, there were a few that were not marked `async`. Everything still 'worked' but it was logging an exception about awaiting something that wasn't awaitable.


ol-iver:
thanks 😄

PR #240, opened 2023-01-23
Do not install global-scope `tests` package

mgorny:
Modify `setup.py` to prevent it from installing the `tests` directory as a global-scope Python package.


ol-iver:
thanks 😃

mgorny:
Thank you!

PR #241, opened 2023-01-24
Add sound modes from telnet connection

starkillerOG:
When using the new telnet interface, the sound modes are reported diffrently on my Marantz SR5008.
This adds the sound modes as they are reported by the telnet connection.


starkillerOG:
@ol-iver could you merge this?

PR #242, opened 2023-01-25
Update telnet code to use readuntil instead of building a buffer.

dcmeglio:
Per some feedback on the HA integration update PR, this converts the telnet code to use readuntil instead of building our own buffers to improve performance and reduce code complexity.

### DIFF
--- a/denonavr/api.py
+++ b/denonavr/api.py
@@ -34,8 +34,6 @@ from .const import (
 
 _LOGGER = logging.getLogger(__name__)
 
-_SOCKET_READ_SIZE = 135
-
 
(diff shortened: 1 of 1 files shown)


bdraco:
https://github.com/ol-iver/denonavr/compare/main...bdraco:denonavr:refactor_protocol

This branch made it stable for me

Note: do not consider this a complete implementation.  I only did enough work to get it stable for me.  It needs some testing. I did not do any reconnect testing or testing to see what happens if the receiver is up but telnet fails to connect. 

Feel free to use all or none of it.

dcmeglio:
No longer needed due to https://github.com/ol-iver/denonavr/pull/243

PR #243, opened 2023-01-29
Telnet Asyncio.Protocol

dcmeglio:
I took the work started by @bdraco, added unit tests, linter cleanups, a couple bug fixes/improvements. I tested this under several reconnect scenarios:

1. Yank the ethernet from the receiver -- properly reconnects fairly instantaneously
2. Disabled wifi on the laptop where I'm running the software -- properly reconnects but only after 5 minutes when the receiver finally realizes the socket dropped and closes the connection
3. Using my unifi router to force disconnect my receiver -- properly reconnects but only after 5 minutes when the receiver finally realizes the socket dropped and closes the connection

Changes from bdraco's original branch:
1. Changed the "ping" command from `PW?` to `SLP?` -- this prevents callbacks from triggering when we get the response to a PW command, since we don't care about SLP
2. Fixed a small issue with the reconnect logic where it wasn't checking if it was healthy and I had 1 time where the reconnect logic seemed to continue even after the reconnect succeeded so I added a check to bail out if it's healthy

If you do decide to merge this, definitely add @bdraco as a contributor, he did the bulk of the work!


dcmeglio:
@ol-iver Seems pylint decided to add something else they don't like so I fixed that in this PR as well to fix your failing nightly checks. Still would be great to get you to review this so we can update HA.

starkillerOG:
@ol-iver I am also eagerly awaiting this functionallity, would be greath if you could find some time to review/merge :)

denonavr/api.py
```diff
@@ -11,8 +11,10 @@
 import logging
 import xml.etree.ElementTree as ET
 
+import time
 from io import BytesIO
-from typing import Awaitable, Callable, Dict, Hashable, Optional, Tuple
+from typing import cast, Awaitable, Callable, Dict, Hashable, Optional, Tuple
+import async_timeout
```
ol-iver: Please remove `async_timeout`.
This is an external dependency used for one single call only. Btw. it was not mentioned in `setup.py` too.

denonavr/api.py
```diff
@@ -11,8 +11,10 @@
 import logging
 import xml.etree.ElementTree as ET
 
+import time
```
ol-iver:
```suggestion
import logging
import time
import xml.etree.ElementTree as ET

```

denonavr/api.py
```diff
@@ -316,134 +318,210 @@ def is_default_async_client(self) -> bool:
...
+        try:
+            async with async_timeout.timeout(self.timeout):
+                transport_protocol = await loop.create_connection(
+                    lambda: DenonAVRTelnetProtocol(
+                        on_connection_lost=self._disconnected,
+                        on_message=self._process_event
+                    ),
+                    self.host,
+                    23
+                )
```
ol-iver:
```suggestion
            transport_protocol = await asyncio.wait_for(loop.create_connection(
                lambda: DenonAVRTelnetProtocol(
                    on_connection_lost=self._disconnected,
                    on_message=self._process_event
                ),
                self.host,
                23
            ), timeout=self.timeout)
```
~~As mentioned before, please remove `async_timeout` dependency. It is easy to replace this call by standard python expressions.~~

requirements.txt
```diff
@@ -3,3 +3,4 @@ attrs>=21.2.0
 defusedxml>=0.7.1
 httpx>=0.21.0
 netifaces>=0.11.0
+async-timeout>=4.0.2
```
ol-iver:
```suggestion
async-timeout>=4.0.2 ; python_version < "3.11"
```
and please add the same to `install_requires` in `setup.py`

denonavr/api.py
```diff
@@ -316,134 +318,210 @@ def is_default_async_client(self) -> bool:
...
+                self._protocol.close()
+            self._disconnected()
+            return
+
+        if time_since_response > _MONITOR_INTERVAL and self._protocol:
+            # Keep the connection alive
+            _LOGGER.debug("%s: Sending keep alive", self.host)
+
+            # Use a command that won't trigger any callbacks
+            self._protocol.write("SLP?\r")
```
ol-iver: I don't see any advantage in this. Processing the `power` callback is not expensive. On the contrary, this command might not be supported by all receivers.
```suggestion
            self._protocol.write("PW?\r")
```

denonavr/api.py
```diff
@@ -316,134 +318,210 @@ def is_default_async_client(self) -> bool:
...
+        """Monitor the connection."""
+        time_since_response = time.monotonic() - self._last_message_time
+        if time_since_response > _MONITOR_INTERVAL * 2:
+            _LOGGER.info(
+                "%s: Keep alive failed, disconnecting and reconnecting",
+                self.host
+            )
+            if self._protocol:
+                self._protocol.close()
+            self._disconnected()
```
ol-iver:
```suggestion
            self._handle_disconnected()
```

denonavr/api.py
```diff
@@ -316,134 +318,210 @@ def is_default_async_client(self) -> bool:
...
+
+        if time_since_response > _MONITOR_INTERVAL and self._protocol:
+            # Keep the connection alive
+            _LOGGER.debug("%s: Sending keep alive", self.host)
+
+            # Use a command that won't trigger any callbacks
+            self._protocol.write("SLP?\r")
+        self._schedule_monitor()
+
+    def _disconnected(self) -> None:
```
ol-iver:
```suggestion
    def _handle_disconnected(self) -> None:
```

ol-iver (commented):
Thank you @dcmeglio
Found a couple of small things. Otherwise, it looks good 🚀

denonavr/api.py
bdraco: In core, we prefer `async_timeout` as it doesn't create a task which reduces the overhead.  It will be an asyncio built-in in python 3.11+

```
if sys.version_info[:2] < (3, 11):
    from async_timeout import timeout as asyncio_timeout
else:
    from asyncio import timeout as asyncio_timeout
```

bdraco: Some additional background: we have slowly been replacing asyncio.wait_for in core because there were so many tasks being created from all the various uses by integrations that it led to increased event loop latency and made HA less responsive. We now include measuring the number of tasks for minute when we get performance issues opened for core from profiles. This effort (combined with other efforts to reduce tasks) has been largely successful the average number of tasks per minute has dropped significantly and we have a lot fewer latency complaints these days.

ol-iver: okay, then let's try it like this 😄 

```suggestion
if sys.version_info[:2] < (3, 11):
    from async_timeout import timeout as asyncio_timeout
else:
    from asyncio import timeout as asyncio_timeout
```

denonavr/api.py
ol-iver: Not relevant anymore

denonavr/api.py
```diff
@@ -316,134 +318,210 @@ def is_default_async_client(self) -> bool:
...
+                        self.host
+                    )
+                except AvrNetworkError as ex:
+                    _LOGGER.debug("%s: %s", self.host, ex, exc_info=True)
+                except Exception:    # pylint: disable=broad-except
+                    _LOGGER.error(
+                        "%s: Unexpected exception on telnet reconnect",
+                        self.host,
+                        exc_info=True
+                    )
```
ol-iver: Please switch the loop and the lock. Otherwise, `async_disconnect` would wait forever while it is reconnecting.
```suggestion
        while self._connection_enabled is True and not self.healthy:
            async with self._connect_lock:
                try:
                    await self._async_establish_connection()
                except AvrTimoutError:
                    _LOGGER.debug(
                        "%s: Timeout exception on telnet reconnect",
                        self.host
                    )
                except AvrNetworkError as ex:
                    _LOGGER.debug("%s: %s", self.host, ex, exc_info=True)
                except Exception:    # pylint: disable=broad-except
                    _LOGGER.error(
                        "%s: Unexpected exception on telnet reconnect",
                        self.host,
                        exc_info=True
                    )
```

denonavr/api.py
```diff
@@ -316,134 +318,210 @@ def is_default_async_client(self) -> bool:
...
-                    data = bytearray()
+                    _LOGGER.info("%s: Telnet reconnected", self.host)
+                    return
 
-        if self.connected is True:
-            _LOGGER.info(
-                "Telnet connection terminated by receiver, reconnecting")
-            self._monitor_task = asyncio.create_task(self._async_reconnect())
+                await asyncio.sleep(backoff)
+                backoff = min(30.0, backoff*2)
```
ol-iver: This should be outside the lock. See previous comment.
```suggestion
            await asyncio.sleep(backoff)
            backoff = min(30.0, backoff*2)
```

ol-iver (commented):
Found one more thing in the reconnection flow.

denonavr/api.py
dcmeglio: Resolved in https://github.com/dcmeglio/denonavr/commit/2362b017c3c80bc993a3e753f246fcb4665cea62

denonavr/api.py
dcmeglio: Resolved in https://github.com/dcmeglio/denonavr/commit/2362b017c3c80bc993a3e753f246fcb4665cea62

denonavr/api.py
dcmeglio: Resolved in https://github.com/dcmeglio/denonavr/commit/2362b017c3c80bc993a3e753f246fcb4665cea62

requirements.txt
dcmeglio: Resolved both the requirements and setup in https://github.com/dcmeglio/denonavr/commit/b5ab97ff9b19dfb8ec661e426694b372cb154449

denonavr/api.py
dcmeglio: Resolved in https://github.com/dcmeglio/denonavr/commit/2362b017c3c80bc993a3e753f246fcb4665cea62

denonavr/api.py
dcmeglio: Good catch! Resolved in https://github.com/dcmeglio/denonavr/commit/c4ef1fcd87ed5c51313699dde6f5d6ca46c40a2b

denonavr/api.py
dcmeglio: Resolved in https://github.com/dcmeglio/denonavr/commit/c4ef1fcd87ed5c51313699dde6f5d6ca46c40a2b

denonavr/api.py
dcmeglio: This is an easy change if you feel strongly about it. My rationale was this - agreed, triggering a callback is inexpensive. However, what this means is the consumer will receive a message indicating a state change every 30 seconds. So in the case of HA, it'll mean triggering a state change and all that good stuff. Just seemed like overhead. The SLP command was added to the Denon protocol at least as far back as 2009. https://assets.denon.com/DocumentMaster/UK/436_AVR4810CI_4810_PROTOCOL_V660a.pdf but it did NOT exist back in 2006 http://assets.eu.denon.com/DocumentMaster/FR/AVR4306_PROTOCOL_V4.6a.pdf  

Denon doesn't seem to provide a list of which models support the timer, without me going through model by model (see https://support-uk.denon.com/app/answers/detail/a_id/1173/~/which-denon-devices-provide-timer-functions%3F)

So I changed this back to PW? per your request in https://github.com/dcmeglio/denonavr/commit/5c5aa1cdcf95be912318d593085b9c2abecab4e3

dcmeglio:
All comments addressed and resolved merge conflicts.

ol-iver:
looks good, thank you 😄

PR #244, opened 2023-02-02
Update SOUND_MODE_MAPPING for Marantz NR-1602

edwardsre:
These values were observed viewing the entity state in Home Assistant for my Marantz NR-1602 receiver. I would like to add these values to complete the sound mode mapping.

One example of a value being added. Currently, all three new values map to the sound mode `DOLBY DIGITAL` even though I don't see any mappings to that sound mode for those values.

image

All other mapping seem to work.


ol-iver:
Thank you @edwardsre 
I just converted your sound modes to upper case. That's how it is done for all sound modes.

PR #245, opened 2023-02-17
Disconnect fix

dcmeglio:
Fixes a bug due to _reconnect_task already being None


denonavr/api.py
```diff
@@ -493,11 +493,6 @@ async def async_disconnect(self) -> None:
                 self._protocol.close()
                 self._protocol = None
 
-            try:
-                await self._reconnect_task
```
dcmeglio, line 497: This is set to None on line 491 causing this to always throw an error.

bdraco: We should still await the task in case there was an exception to avoid an unretrieved exception error.  Save it in a local var before its set to none and await it only if it exists (be sure to guard against canceled error in case it was canceled so the cancel error does not raise up and cancel this coro).

dcmeglio: Ah now I get what you were trying to achieve. Thanks for the explanation. I think this will take care of it.

bdraco: It won't make much difference but I'd put the if check around the try instead of in the try since we usually only want code that can fail inside the try.  Also its more efficient as it avoids setting up a try if we don't need it.

bdraco (approved):

ol-iver:
@dcmeglio please start your fix from a clean branch. This PR includes 20 commits which are part of your previous PR.

dcmeglio:
Oops yup. I’ll fix it later today

PR #246, opened 2023-02-17
Fix an issue with reconnect_task being None.

dcmeglio:
Resolve an issue where reconnect_task was set to None but still being used causing async_disconnect to fail.


ol-iver:
thanks 😄

PR #253, opened 2023-04-16
Use telnet commands to interact with the receiver when the connection is healthy

ol-iver:
With this PR telnet commands are used to interact with the receiver when telnet is connected and the connection is healthy. Otherwise, HTTP commands are used.

@dcmeglio @bdraco @starkillerOG If you would like to test this, please feel invited to do so 😄


bdraco:
not stale, awaits testing

PR #264, opened 2023-10-03
Improve 169.254.0.0/16 SSDP broadcast avoidance check

scop:
It was unnecessarily using a regular expression, and in suspiciously glob-style manner.


ol-iver:
Awesome, thanks 😄

Issue #289, opened 2024-03-12
Add remote control / system control capabilities


klada:
Right now there is no button navigation (cursor up/down/left/right/...) available through this library. This would however be very handy so the device could be controlled through IP entirely - no more IR remote would be required for navigating the setup menu.

Denon AVRs do support the following actions through the Telnet interface:

- Cursor Up: `MNCUP`
- Cursor Down: `MNCDN`
- Cursor Left: `MNCLT`
- Cursor Right: `MNCRT`
- Cursor Enter: `MNENT`
- Info button: `MNINF`
- Option button: `MNOPT`
- Back button: `MNRTN`
- Open setup menu: `MNMEN ON`
- Close setup menu: `MNMEN OFF`
- Get menu status: `MNMEN?` → Returns either `MNMEN ON` or `MNMEN OFF`.

HTTP control using these commands works as follows:

```
# Move cursor up
GET http://IP:8080/goform/formiPhoneAppDirect.xml?MNCUP
# Open setup menu
GET http://IP:8080/goform/formiPhoneAppDirect.xml?MNMEN%20ON
```

Note on the setup menu: It seems like the state can only be queried through Telnet (`MNMEN?` command). The phone app apparently maintains the setup menu's state (open or closed) internally and does not issue a `MNMEN?` query in order to toggle it.

albaintor:
Hi, done in PR  https://github.com/ol-iver/denonavr/pull/290

PR #290, opened 2024-03-16
Added additional commands

albaintor:
Added additional commands for AVR : cursor pad, settings and option menu, info as requested in https://github.com/ol-iver/denonavr/issues/289


klada:
I think you are always sending HTTP commands, even if the library is in telnet control mode. Or did I miss something here?

albaintor:
> I think you are always sending HTTP commands, even if the library is in telnet control mode. Or did I miss something here?

Indeed I am not aware (yet) on how the library works. I added only HTTP commands. Is it the same command names for telnet ?
Also one of the commands expects a reply, is it possible to get a response for telnet commands ?
I will check after this anyway

albaintor:
Hi again, I added corresponding commands for telnet protocole according to this documentation
https://www.heimkinoraum.de/upload/files/product/IP_Protocol_AVR-Xx100.pdf

albaintor:
H @ol-iver, do you intend to merge the PR soon ?
Thank you

ol-iver:
I like the idea, thanks for your PR 🚀  
This is a private project, so I don't have always enough time for it.
However, I'll check your PR this weekend 🙂

ol-iver:
I just enabled the test. You could already start fixing the tests and the linting issues.
In case you want to run them locally, please check the github workflows how you can run [linting](https://github.com/ol-iver/denonavr/blob/main/.github/workflows/python-code-format.yml) and [unit](https://github.com/ol-iver/denonavr/blob/main/.github/workflows/python-tests.yml) tests.

albaintor:
Hi, I have updated formatting

denonavr/const.py
```diff
@@ -375,6 +408,17 @@
...
+    command_cusor_down=COMMAND_CURSOR_DOWN,
+    command_cusor_left=COMMAND_CURSOR_LEFT,
+    command_cusor_right=COMMAND_CURSOR_RIGHT,
+    command_cusor_enter=COMMAND_CURSOR_ENTER,
+    command_back=COMMAND_BACK,
+    command_info=COMMAND_INFO,
+    command_options=COMMAND_OPTIONS,
+    command_setup_open=COMMAND_SETUP_OPEN,
+    command_setup_close=COMMAND_SETUP_CLOSE,
+    command_setup_query=COMMAND_SETUP_QUERY
```
ol-iver: These lines should be added to `ZONE2_URLS` and `ZONE3_URLS` too.
This might be the reason why the tests are failing.

denonavr/foundation.py
```diff
@@ -535,6 +535,87 @@ async def async_power_off(self) -> None:
         else:
             await self.api.async_get_command(self.urls.command_power_standby)
 
+    async def async_cursor_up(self) -> None:
```
ol-iver, line 538: Please call each of your new functions from `DenonAVR` class like in [this example](https://github.com/ol-iver/denonavr/blob/2f18b79ea9cda6ea5c6310300e3767d76e42ec2f/denonavr/denonavr.py#L667-L669). Otherwise the methods won't be publicly accessible.

ol-iver (changes requested):
Found two things.
The first finding might fix the failing tests.

albaintor:
Hi, I have added the missing commands (unit testings passed this time) and added public methods

ol-iver:
Thanks 🙂 
There is one last thing. You too many methods. Those decorated with ` @run_async_synchronously` are deprecated and I removed them with https://github.com/ol-iver/denonavr/pull/291 for all other methods.

albaintor:
Hi @ol-iver, I have applied the merge. It should be okay this time :)

ol-iver:
There are still a couple of linting issues 😅 
Unless you contributed for the first time, Github does not start the tests automatically. Thus is would make sense if you run the [code-format](https://github.com/ol-iver/denonavr/blob/c70e8ce11cd1f9ef86053201588d41d1f63b6e14/.github/workflows/python-code-format.yml#L31-L42) and the [integration tests](https://github.com/ol-iver/denonavr/blob/c70e8ce11cd1f9ef86053201588d41d1f63b6e14/.github/workflows/python-tests.yml#L34-L36) locally.

albaintor:
Hi,

I missed one of the formatting tasks. It should be (hopefully) ok this time

I have a question : are you sure that telnet is faster than HTTP commands ? It may be my setup but I find volume up/down more reactive with http commands

Thanks
Damien

De : Oliver ***@***.***>
Envoyé : jeudi 2 mai 2024 10:12
À : ol-iver/denonavr ***@***.***>
Cc : albaintor ***@***.***>; Author ***@***.***>
Objet : Re: [ol-iver/denonavr] Added additional commands (PR #290)


There are still a couple of linting issues 😅
Unless you contributed for the first time, Github does not start the tests automatically. Thus is would make sense if you run the code-format<https://github.com/ol-iver/denonavr/blob/c70e8ce11cd1f9ef86053201588d41d1f63b6e14/.github/workflows/python-code-format.yml#L31-L42> and the integration tests<https://github.com/ol-iver/denonavr/blob/c70e8ce11cd1f9ef86053201588d41d1f63b6e14/.github/workflows/python-tests.yml#L34-L36> locally.

ol-iver:
There was still an issue. I fixed the last one by myself. 
Please run the tests locally the next time 😉 

The main advantage of telnet is that the properties of your receiver are updated immediately, even when you are pressing buttons on your remote control. When you use HTTP you have to poll for updates with `async_update` method.

ol-iver:
Thanks for your work 🚀

PR #322, opened 2025-01-25
Add global option to skip confirmation when using Telnet

### LINKED ISSUE #321: Slow command execution when using Telnet on AVC-A10H
I've noticed that whenever I execute a command that doesn't use `skip_confirmation=True` each command takes 3 seconds. Using HTTP doesn't have this problem, and I'm wondering what the best approach to fix this would be (I'd like to provide a PR to fix it).

I've been thinking of two solutions:
1. Property `skip_confirmation` in `DenonAVRTelnetApi` with default value `False`
2. Expose the `skip_confirmation` flag on all commands with default value `False`
3. Expose flag to force use HTTP even when Telnet is available

Option 1 is probably easiest to use, but doesn't give any granular control, and might cause issues with the existing overload on the send command methods. Option 3 feels unintuitive.

henrikwidlund:
Fixes #321


henrikwidlund:
Library not at fault

PR #323, opened 2025-03-09
Add support for more commands

henrikwidlund:
Adds support for the following:

__Audyssey__
- LFC
- Containment Amount

__DenonAVR__
- Improved Settings Menu toggle
- Dimmer
- Auto Standby
- Sleep
- Delay and Delay Time
- Eco Mode
- HDMI Output
- HDMI Audio Decode
- Video Processing Mode
- Tactile Transducer
- Room Size
- Triggers
- Speaker Preset
- Bluetooth Transmitter and Output Mode
- Audio Restorer
- Status Display (Denon only)
- System Reset
- Network Restart
- Remote Control and Panel Lock
- Quick Select and Quick Select Memory (programs a Quick Select)

__Dirac__
- Dirac Filter

__Sound Mode__
- Sound Mode Next/Previous
- Neural:X Toggle
- IMAX Toggle
- IMAX Audio Settings Toggle
- IMAX Low and High Pass Filter
- IMAX Subwoofer Output Mode
- Cinema EQ Toggle
- Center Spread Toggle
- Loudness Management Toggle
- Dialog Enhancer
- Auro-Matic 3D Preset
- Auro-Matic 3D Strength
- Auro 3D mode
- Dialog Control
- Speaker Virtualizer
- Effect Speaker Selection

__Volume__
- Channel Volumes
- Reset All Channel Volumes
- Subwoofer On/Off
- Subwoofer Level
- LFE Volume
- Bass Sync


henrikwidlund:
@ol-iver what should I do about the problems with `Literal` in the tests? Replace them with `Enum`?

ol-iver:
> @ol-iver what should I do about the problems with `Literal` in the tests? Replace them with `Enum`?

I'll remove support for python 3.7. It is EOL anyway.

henrikwidlund:
Damnit, I missed your comment and now I pushed the changes. Do you want me to revert it?

henrikwidlund:
If anyone had any doubts, the number of commits to fix linting should tell you that I'm not a Python developer 😅

ol-iver:
> Damnit, I missed your comment and now I pushed the changes. Do you want me to revert it?

Please give it a try 👍 
I just removed Python 3.7 in main branch.

henrikwidlund:
> > Damnit, I missed your comment and now I pushed the changes. Do you want me to revert it?
> 
> Please give it a try 👍 I just removed Python 3.7 in main branch.

Done. Just realized that I could create literals in the const file to not have to duplicate the definitions, that would also work for the channel enum, but I'm not sure about that one as the channel names are a bit hard to understand the acronyms, that's the reason I chose to use enums for those two types, `DimmerMode` is kind of understandable I guess.

henrikwidlund:
Okay, I converted everything to literals, still not sure about the channels though.

henrikwidlund:
@ol-iver okay, I think I'm done with this one now 😅
I looked into having update functionality for the Dirac filter as well as map the names of the profiles, but it's not available in the app, so not sure it's possible yet, ot least it will be very difficult to reverse engineer it. The command  /goform/formiPhoneAppDirect.xml?PSDIRAC%20? doesn't return anything

denonavr/audyssey.py
```diff
@@ -94,6 +97,10 @@ async def _async_sound_detail_callback(
             self._dynamiceq = "1"
         elif parameter == "DYNEQ OFF":
             self._dynamiceq = "0"
+        elif parameter == "PSLFC ON":
```
ol-iver:
```suggestion
        elif parameter == "LFC ON":
```
The first 2 characters of the telnet response form the `event` the remaining ones the `parameter`  (see [here](https://github.com/ol-iver/denonavr/blob/0e4f0338a6865759db7c9b110a2e59a1e1fb0dd9/denonavr/api.py#L712-L715)).

denonavr/audyssey.py
```diff
@@ -94,6 +97,10 @@ async def _async_sound_detail_callback(
...
+            self._lfc = "1"
+        elif parameter == "PSLFC OFF":
```
ol-iver:
```suggestion
        elif parameter == "LFC OFF":
```

denonavr/audyssey.py
```diff
@@ -60,6 +60,9 @@ class DenonAVRAudyssey(DenonAVRFoundation):
     _multeq_control: Optional[bool] = attr.ib(
         converter=attr.converters.optional(convert_string_int_bool), default=None
     )
+    _lfc: Optional[bool] = attr.ib(
+        converter=attr.converters.optional(convert_string_int_bool), default=None
+    )
```
ol-iver, line 65: Please create a property for lfc.

ol-iver (commented):
Awesome 😄 

WDYT about adding properties and register telnet callbacks for things like `dimmer`, `channel_level`, `delay`, `eco_mode`, `hdmi`  and maybe `dirac` (I don't know if it returns anything). They work for the telnet api only, but could still be interesting.

denonavr/audyssey.py
henrikwidlund: Right, I learned this after writing this code in my settings state pr :)

henrikwidlund:
> Awesome 😄
> 
> WDYT about adding properties and register telnet callbacks for things like `dimmer`, `channel_level`, `delay`, `eco_mode`, `hdmi` and maybe `dirac` (I don't know if it returns anything). They work for the telnet api only, but could still be interesting.

That's a good idea! Couldn't make sense of how the responses would be returned for `?`, the HTTP API only returns 200 and there's nothing coming back from Telnet directly.

henrikwidlund:
> > Awesome 😄
> > WDYT about adding properties and register telnet callbacks for things like `dimmer`, `channel_level`, `delay`, `eco_mode`, `hdmi` and maybe `dirac` (I don't know if it returns anything). They work for the telnet api only, but could still be interesting.
> 
> That's a good idea! Couldn't make sense of how the responses would be returned for `?`, the HTTP API only returns 200 and there's nothing coming back from Telnet directly.

@ol-iver how would I register a callback when the command doesn't seem to belong to a zone? Dimmer for example returns this `DIM OFF`.

_edit:_ Found out how to do it

henrikwidlund:
@ol-iver done 😊

henrikwidlund:
I found some issues with the naming of methods for channel levels, they should be channel volumes. I also want to move them to the volume class as well as add a few more commands, so please hold off on merging this.

henrikwidlund:
@ol-iver okay, now I'm done. I've implemented all the commands in the specification for AVC-A10H except:
- A/B Speaker toggle (switch off front right/left channel)
- Effect (don't know what this even is)
- Zone 2 and 3 ( I don't use them, so can't confirm if it'd work. Should be easy for someone else to implement if requested)
  -  Quick Select
  - Auto Standby
  - Stereo/Mono
  - Front levels
  - High Pass Filter
  - Sleep
  - HDMI Audio

henrikwidlund:
Okay, realized I could use bool instead of returning ON/OFF, now I'm out of ideas :)

henrikwidlund:
@ol-iver I have an idea for how to get some of the toggle commands working via http. The api supports sending IR commands to it by sending the identifiers for the IR command. There isn't an IR command for everything, but the settings and other stuff should work. I'm not going to do this a part of this PR because it's already bigger than what I would consider ok (sorry about that).

Edit: They still don't toggle, only the mute command does, and that one I think is already working with HTTP?

ol-iver (approved):
That's a crazy amount of commands and properties, awesome 🚀 

Are you using home-assistant? Maybe they could be added as services/additional entities. I'm not sure what's the best way there though since I did not do any bigger contributions lately.

ol-iver:
> Edit: They still don't toggle, only the mute command does, and that one I think is already working with HTTP?

IIRC the mute command does not toggle perfectly with HTTP since there must be an `async_update` to get the new state that you can toggle it the other way again.
But yes, this PR is already big enough 😅

ol-iver:
Btw., I'm sorry that it took so long. I did not spend so much time here recently.

henrikwidlund:
I don't use home assistant, I use this on my remote (Unfolded Circle).

No worries with the delay, we all have work and private life that needs to take priority first

ol-iver:
The remote looks good, I'm gonna check it out.

I'm gonna cut a new release soon, that you can use your work here 😄

henrikwidlund:
Awesome, thanks! Once released I'll do a PR to the Denon integration driver with the changes :)

PR #324, opened 2025-03-09
Add support for more sound modes

henrikwidlund:
Sound modes fetched from the AVC-A10H specification


denonavr/const.py
```diff
@@ -276,6 +276,9 @@
         "MULTI IN + NEURAL:X",
         "NEURAL",
         "STANDARD(DOLBY)",
+        "DOLBY AUDIO-DD+NEURAL:X",
+        "DOLBY AUDIO-DD+",
+        "DOLBY AUDIO-DD+ +NERUAL:X",
```
ol-iver, line 282:
```suggestion
        "DOLBY AUDIO-DD+ +NEURAL:X",
```
This is a typo, isn't it? 😅

henrikwidlund: I copied it from the Excel spread sheet that I got from Denon, so I'm not sure actually, I'd hope it's a spelling issue in their docs and not in the firmware. Not sure what to do here to be honest.

ol-iver: We could add both versions 😉

ol-iver (approved):
Thanks 😄

PR #325, opened 2025-03-13
Skip confirmation for commands which don't return anything

### LINKED ISSUE #321: Slow command execution when using Telnet on AVC-A10H

henrikwidlund:
Fixes #321

### DIFF
--- a/denonavr/foundation.py
+++ b/denonavr/foundation.py
@@ -552,7 +552,8 @@ class DenonAVRDeviceInfo:
         """Cursor Up on receiver via HTTP get command."""
         if self.telnet_available:
             await self.telnet_api.async_send_commands(
-                self.telnet_commands.command_cusor_up
+                self.telnet_commands.command_cusor_up,
+                skip_confirmation=True
             )
         else:
             await self.api.async_get_command(self.urls.command_cusor_up)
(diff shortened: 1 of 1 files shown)


ol-iver (changes requested):
Thanks 🙂 

Please skip the confirmation in `async_back` to ([ref](https://github.com/ol-iver/denonavr/pull/325/files#diff-11b5022a6cfe66073ff7af1648b23086f16959c65da9a5937402f1419b1efefcR606)). It is sending `MNRTN` which is also mentioned to be command only.

Not all of your changes are formatted with [black](https://github.com/psf/black).

henrikwidlund:
Just added it for `async_back` and fixed the formatting (I hope).

ol-iver (approved):
Awesome 🚀

PR #328, opened 2025-03-18
Add settings menu state

henrikwidlund:
Allows for settings menu toggle without relying on HTTP. Especially useful if the HTTP request doesn't return anything.


denonavr/foundation.py
```diff
@@ -622,6 +648,18 @@ async def async_options(self) -> None:
 
     async def async_settings_menu(self) -> None:
         """Options menu on receiver via HTTP get command."""
+        # fast path if we already know the state
+        if self.telnet_available and self._settings_menu is not None:
```
ol-iver:
```suggestion
        if self.telnet_available:
```
This looks for me like the telnet path instead of the fast path.
You could add a `MNMEN?` command [here](https://github.com/ol-iver/denonavr/blob/0e4f0338a6865759db7c9b110a2e59a1e1fb0dd9/denonavr/api.py#L569) which make the library query the state each time telnet connects.

Then, you could also remove the telnet handling in the slow path.

denonavr/foundation.py
```diff
@@ -521,6 +538,15 @@ def power(self) -> Optional[str]:
         """
         return self._power
 
+    @property
+    def settings_menu(self) -> Optional[str]:
+        """
+        Return the settings menu state of the device.
+
+        Possible values are: "ON" and "OFF"
```
ol-iver, line 546: Please add to the comment that this function only returns a value when telnet is enabled.

denonavr/denonavr.py
```diff
@@ -278,6 +278,15 @@ def power(self) -> Optional[str]:
...
         return self._device.power
 
+    @property
+    def settings_menu(self) -> Optional[str]:
+        """
+        Return the settings menu state of the device.
+
+        Possible values are: "ON" and "OFF"
+        """
+        return self._device.settings_menu
```
ol-iver, line 288: Please add to the comment that this function only returns a value when telnet is enabled.

ol-iver (changes requested):
Thanks again 🚀

denonavr/foundation.py
henrikwidlund: I hope I understood you correctly with my adjustments :)

denonavr/foundation.py
```diff
@@ -622,21 +648,28 @@ async def async_options(self) -> None:
...
         """Options menu on receiver via HTTP get command."""
-        res = await self.api.async_get_command(self.urls.command_setup_query)
+        # fast path if we already know the state
         if self.telnet_available:
-            if res is not None and res == "MNMEN ON":
+            if self._settings_menu is None:
+                self._settings_menu = await self.api.async_get_command(
+                    self.urls.command_setup_query
+                )
+
```
ol-iver: Almost 😉 

```suggestion
```
You could entirely remove this part and add a `MNMEN?` line after [this line](https://github.com/ol-iver/denonavr/blob/0e4f0338a6865759db7c9b110a2e59a1e1fb0dd9/denonavr/api.py#L569).
This command will be sent when the telnet connection has been established. The receiver will answer with either `MNMEN ON` or `MNMEN OFF` (see the api document I linked before on page 27), so the callback will be called and  `self._settings_menu` won't be None.

henrikwidlund: Aha, I was wondering how the answers would work on Telnet :)

ol-iver (approved):
Nice 👍

PR #330, opened 2025-04-08
Set timeout on _send_confirmation_timeout

henrikwidlund:
The timeout for acknowledgements via Telnet commands will now use the value specified in  the `timeout` parameter in the `DenonAVR` class,


ol-iver (approved):
Thanks, one more time 😄

PR #331, opened 2025-05-01
Fix lookup maps which might cause telnet callback exceptions

ol-iver:
While testing I found two error messages in my logs. This PR should fix the issue.

```
2025-05-01 19:45:08,901 - denonavr.api - DEBUG - Incoming Telnet message: PSSWL OFF
2025-05-01 19:45:08,901 - denonavr.api - ERROR - 192.168.178.250: Event callback caused an unhandled exception: 'OFF'
```
```
2025-05-01 19:45:09,414 - denonavr.api - DEBUG - Incoming Telnet message: PSSP:SB
2025-05-01 19:45:09,415 - denonavr.api - ERROR - 192.168.178.250: Event callback caused an unhandled exception: 'SB'
2025-05-01 19:45:09,415 - denonavr.api - ERROR - 192.168.178.250: Event callback caused an unhandled exception: 'SB'
2025-05-01 19:45:09,415 - denonavr.api - ERROR - 192.168.178.250: Event callback caused an unhandled exception: 'SB'
```

@henrikwidlund could you have a look please?

### DIFF
--- a/denonavr/const.py
+++ b/denonavr/const.py
@@ -1350,12 +1350,29 @@ ECO_MODE_MAP_LABELS = {value: key for key, value in ECO_MODE_MAP.items()}
 EcoModes = Literal["On", "Auto", "Off"]
 """Eco Modes."""
 
-EffectSpeakers = Literal["Floor", "Height + Floor"]
+EffectSpeakers = Literal[
+    "Floor",
+    "Front",
+    "Front Height",
+    "Front Wide",
+    "Front Height + Front Wide",
+    "Height + Floor",
+    "Surround Back",
+    "Surround Back + Front Height",
+    "Surround Back + Front Wide",
+]
 """Effect Speakers."""
 
 EFFECT_SPEAKER_SELECTION_MAP = {
     "Floor": "FL",
+    "Front": "FR",
+    "Front Height": "FH",
+    "Front Wide": "FW",
+    "Front Height + Front Wide": "HW",
     "Height + Floor": "HF",
+    "Surround Back": "SB",
+    "Surround Back + Front Height": "BH",
+    "Surround Back + Front Wide": "BW",
 }
 EFFECT_SPEAKER_SELECTION_MAP_LABELS = {
     value: key for key, value in EFFECT_SPEAKER_SELECTION_MAP.items()
--- a/denonavr/volume.py
+++ b/denonavr/volume.py
@@ -60,9 +60,8 @@ class DenonAVRVolume(DenonAVRFoundation):
     _subwoofer: Optional[bool] = attr.ib(
         converter=attr.converters.optional(convert_on_off_bool), default=None
     )
-    _subwoofer_levels: Optional[Dict[Subwoofers, Union[float, bool]]] = attr.ib(
-        default=None
-    )
+    _subwoofer_levels_adjustment: bool = attr.ib(default=True)
+    _subwoofer_levels: Optional[Dict[Subwoofers, float]] = attr.ib(default=None)
     _valid_subwoofers = get_args(Subwoofers)
     _lfe: Optional[int] = attr.ib(converter=attr.converters.optional(int), default=None)
     _bass_sync: Optional[int] = attr.ib(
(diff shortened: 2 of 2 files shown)


henrikwidlund:
Interesting, the `PSSWL OFF` isn't documented in the API, but I guess it could be solved by adding `ON` and `OFF` to `CHANNEL_VOLUME_MAP`.

As for `PSSP:SB`, `_speaker_virtualizer` might need to be adjusted to handle string values as well. This is also not documented, all the Excel file I got from Denon says is ON/OFF. I guess `SB` means surround back?

henrikwidlund:
[Denon FY2024_AVR_PROTOCOL_V01.xlsx](https://github.com/user-attachments/files/20002950/Denon.FY2024_AVR_PROTOCOL_V01.xlsx)

ol-iver:
`PSSWL OFF` might mean something else. I did not add it to `CHANNEL_VOLUME_MAP` because I don't know to which numerical value it could be mapped. Additionally, the next message my receiver sends is `PSSWL 50` and the level should not flip. 
```
2025-05-01 19:45:08,901 - denonavr.api - DEBUG - Incoming Telnet message: PSSWL OFF
2025-05-01 19:45:08,901 - denonavr.api - ERROR - 192.168.178.250: Event callback caused an unhandled exception: 'OFF'
Zone: Main Event: PS Parameter: SWL OFF
2025-05-01 19:45:08,938 - denonavr.api - DEBUG - Incoming Telnet message: PSSWL 50
Zone: Main Event: PS Parameter: SWL 50
```

Yes `SB` means surround back.

I can't say anything about `_speaker_virtualizer`. My receiver does not send a response to `PSSPV ?`.

henrikwidlund:
Okay, I'll look into it tomorrow. I didn't mean to map it to a numerical value, I meant to have a union of on/off and numerical value

ol-iver:
In the document you attached you can find that `PSSWL ON` (and probably `PSSWL OFF` too) is part of the response to `PSSWL ?` command.
```
"PSSWL ?<CR>
PSSWL ON<CR>
PSSWL 50<CR>
PSSWL2 50<CR>
PSSWL3 50<CR>
PSSWL4 50<CR>
※"PSSWL2",  "PSSWL3" or "PSSW4" command is not output if it is not set."
```
These are probably two different attributes. Like an ON/OFF switch and a level.

henrikwidlund:
Yeah, must've missed that. Then I guess the union is the right way forward for that at least

henrikwidlund:
@ol-iver I created another PR to fix the problems you found, also adjusted tactile transducer as the docs doesn't seem accurate there (power state and levels are two different APIs).

#332

henrikwidlund:
No need to adjust speaker virtualizer, it should work as is. I did go through the document again, and this PR and the one you just merged from me should take care of the errors you found. But you never know since they don't supply you with all possible values in the excel file 😕

ol-iver:
I introduced `_subwoofer_levels_adjustment` as a new `volume` property. It contains the `ON/OFF` state.
`_subwoofer_levels` is returned only when `_subwoofer_levels_adjustment`  is True. When I switch the adjustment off, it still returns the last level adjustments I made.
`_subwoofer_levels_adjustment` is True by default, that receivers which do not return `PSSWL ON/OFF` still see the subwoofer levels.

henrikwidlund:
Looks good, not sure if we should expose the new bool for subwoofer as well? Also, is the on/off global, or should it be made to map to specific subwoofers?

ol-iver:
On my receiver it is a global value. I'm able to switch the entire feature on and off. If it is on, I can adjust the level of two subwoofers.

It is already possible to find out whether the feature is ON or OFF. If `_subwoofer_levels` return None, it is off (or not existing). If it returns a dictionary, it is on.
Assuming not all receiver send `PSSWL ON/OFF`, we would need to make `_subwoofer_levels_adjustment` optional as well if we expose it. In my opinion, this would complicate things for only little additional benefit.

henrikwidlund:
Then I'm happy with the solution!

PR #332, opened 2025-05-02
Correct value types based on issues after testing

henrikwidlund:
Adjusts the types for Tactile Transducer and Subwoofer Levels.

### DIFF
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -623,13 +623,11 @@ class DenonAVR(DenonAVRFoundation):
         return self._device.video_processing_mode
 
     @property
-    def tactile_transducer(self) -> Optional[str]:
+    def tactile_transducer(self) -> Optional[bool]:
         """
         Return the tactile transducer state of the device.
 
         Only available if using Telnet.
-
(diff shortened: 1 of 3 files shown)


ol-iver (approved):
Let's give it a try.

ol-iver:
Unfortunately, it does not work correctly. I better should have read my own comment first (https://github.com/ol-iver/denonavr/pull/331#issuecomment-2845418379).
Now the value flips because my receiver always sends two message like `PSSWL OFF` and `PSSWL 50` or `PSSWL ON` PSSWL 50`.

`OFF` does not mean the subwoofer is off, but the level adjustments are off. We should use two attributes to store it.

henrikwidlund:
Oh, okay. I'm not at home atm, but I can fix that in a couple of hours unless you want to do it yourself? Might be better if you do it regardless since my receiver doesn't behave in that way.

Edit: or maybe it does, but since I have Dirac enabled, which takes control of all levels and such, it might not surface for me.

ol-iver:
I updated #331. You can have a look when your back home 🙂

PR #333, opened 2025-05-04
Remove effect speaker toggle

henrikwidlund:
A toggle doesn't make sense now that there are more than two effect speakers


ol-iver (approved):
Makes sense.

PR #337, opened 2025-05-24
Fix incorrect mappings and validators

henrikwidlund:
Fixes problems with mapping and validation of Dirac, Eco and HDMI Output commands.


henrikwidlund:
@ol-iver I went through my previous changes where I added all of the commands and this was the only one that was incorrect.

edit: found two more methods with issues in them.

ol-iver (approved):
Thanks 🙂

PR #344, opened 2025-08-01
Group callbacks by event type to improve performance

henrikwidlund:
Reduce the number of callbacks as well as make them register async to improve startup time on weaker hardware where it can take several seconds.

This reduces the number of callbacks from 49 to 22.

### DIFF
--- a/denonavr/audyssey.py
+++ b/denonavr/audyssey.py
@@ -7,6 +7,7 @@ This module implements the Audyssey settings of Denon AVR receivers.
 :license: MIT, see LICENSE for more details.
 """
 
+import asyncio
 import logging
 from collections.abc import Hashable
 from typing import List, Optional
@@ -71,21 +72,20 @@ class DenonAVRAudyssey(DenonAVRFoundation):
     # AppCommand0300.xml interface
     appcommand0300_attrs = {AppCommands.GetAudyssey: None}
 
-    def setup(self) -> None:
+    async def async_setup(self) -> None:
         """Ensure that the instance is initialized."""
         # Add tags for a potential AppCommand.xml update
         for tag in self.appcommand0300_attrs:
             self._device.api.add_appcommand0300_update_tag(tag)
 
-        self._device.telnet_api.register_callback(
-            "PS", self._async_sound_detail_callback
-        )
+        asyncio.create_task(self._async_register_audyssey_callbacks())
 
         self._is_setup = True
 
-    async def _async_sound_detail_callback(
-        self, zone: str, event: str, parameter: str
-    ) -> None:
+    async def _async_register_audyssey_callbacks(self) -> None:
+        self._device.telnet_api.register_callback("PS", self._async_ps_callback)
+
+    async def _async_ps_callback(self, zone: str, event: str, parameter: str) -> None:
         """Handle a sound detail change event."""
         if self._device.zone != zone:
             return
@@ -114,7 +114,7 @@ class DenonAVRAudyssey(DenonAVRFoundation):
         _LOGGER.debug("Starting Audyssey update")
         # Ensure instance is setup before updating
         if not self._is_setup:
-            self.setup()
+            await self.async_setup()
 
         # Update state
         await self.async_update_audyssey(global_update=global_update, cache_id=cache_id)
--- a/denonavr/denonavr.py
+++ b/denonavr/denonavr.py
@@ -165,12 +165,12 @@ class DenonAVR(DenonAVRFoundation):
                 self._name = self._device.friendly_name
 
             # Setup other functions
-            self.input.setup()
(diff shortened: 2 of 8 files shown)


henrikwidlund:
@ol-iver this probably needs some thorough testing.

Were could reduce this even further by listening on all events in a global handler that will set the properties in the respective classes.

ol-iver:
Same here, I'll have a look after my summer holiday 🏝️

henrikwidlund:
The more I think about it, the more I think that handling all events in one class makes sense. It will have the properties for states and then all other classes will just call those properties in their properties. This would reduce the callbacks to just one and all of that handling would be in one place.

Happy vacationing!

henrikwidlund:
@ol-iver I would like to use switch statements to optimize this further, but that's not available in python 3.9, could support for that be dropped? It's a few months away from EOL anyway :)

henrikwidlund:
@ol-iver gentle bump :)

ol-iver:
Thanks for looking into the performance topic 😄 

The asynchronous processing of callback message might have the biggest performance impact to the telnet API. Performance wise, it was probably not a good idea to create tasks on each event ([ref](https://github.com/ol-iver/denonavr/blob/1f0a373f897ecefa77618218ddd8192c8dd16b93/denonavr/api.py#L800-L802)). This creates some overhead in python and the library processes quite a few events.
I briefly checked the telnet callbacks. They are only setting values and no callback is doing IO stuff. Thus, we could move `_async_run_callbacks` into `_run_callbacks` and all callbacks as well to reduce the load on the event loop.
WDYT?

denonavr/audyssey.py
```diff
@@ -71,21 +72,20 @@ class DenonAVRAudyssey(DenonAVRFoundation):
...
+    async def async_setup(self) -> None:
         """Ensure that the instance is initialized."""
         # Add tags for a potential AppCommand.xml update
         for tag in self.appcommand0300_attrs:
             self._device.api.add_appcommand0300_update_tag(tag)
 
-        self._device.telnet_api.register_callback(
-            "PS", self._async_sound_detail_callback
-        )
+        asyncio.create_task(self._async_register_audyssey_callbacks())
```
ol-iver, line 81: When creating task instead of running function synchronously the overall performance is slower not faster.
You might not notice this here, because this only creates the task for a function and does not wait for its execution. Since the task is referenced nowhere, it is even possible that python garbage collector collects it before it is executed.

henrikwidlund: Yeah, I noticed. I've actually fixed that in my fork, but didn't submit it here, had to make other changes that I don't think you want either. You might want to have a look at the main branch in the fork and see what from it that you'd want.

It's also not the callbacks that causes the slowdown, it's the registration of events on telnet without delays that overwhelms the receiver. Only sending four or five of them locks the receiver up until it has processed everything. I solved this by sorting the registrations by importance and adding a 100ms delay between each registration.

ol-iver: What do you mean with `locks the receiver up`? Does it react slowly, is it ignoring commands or is it something else?

henrikwidlund: It locks it up as in not processing any additional commands until it's done with the event responses, resulting in timeouts. This is perhaps not very noticeable in HA where you rarely reboot, but for the remote where this library is used it causes frustration because the remote goes into sleep mode after a few minutes and then the connection is established again when you for example want to change the volume.

ol-iver: Ok, so it would be good to schedule a task when telnet is connecting and send the commands with a small break as you did in your main branch. This will take 6 seconds or so, but most people would not notice that.
You could improve your implementation a bit though.
-  Save a reference to the task while it is running, that it is not garbage collected too early (see https://docs.python.org/3/library/asyncio-task.html#creating-tasks)
- Cancel the task and remove its reference when telnet is disconnecting while it is still running. This avoids that we flood the system with tasks when the connection is unstable.

Regarding the callbacks I prefer to switch them all to synchronous functions to reduce complexity. We also avoid creating a new task on every callback call. I raised PR #349. PTAL.

When we do all this, I don't think that it makes a real difference regarding performance whether we group the callbacks or not.

henrikwidlund: Sorry for not getting back to you, been busy with all kinds of stuff. I guess this PR can be closed now, seeing that you opened another PR for it?

ol-iver: Yes, I think switching to synchronous callbacks and avoiding that the receiver is blocked by a storm of commands are the most important things performance wise.
Grouping the callbacks additionally might bring only very little performance improvements even on tiny machines.

henrikwidlund:
Superseded by #351

PR #345, opened 2025-08-16
Fix quick/smart select mode/memory mix-up

henrikwidlund:
The mode and memory commands uses mixed up constants when using telnet.


ol-iver (approved):
Thanks 👍

PR #351, opened 2025-10-27
Throttle telnet update commands when reconnecting

ol-iver:
This PR throttles telnet commands which are sent to triggers updates when telnet is reconnecting.

@henrikwidlund I reused your sorting (https://github.com/henrikwidlund/denonavr/commit/786f34f82ba56dc1ddf4d655bf5b686c4bc3587c), but the calls are waiting for confirmation with a low timeout now instead of using a fix delay.

### DIFF
--- a/denonavr/api.py
+++ b/denonavr/api.py
@@ -538,76 +538,88 @@ class DenonAVRTelnetApi:
         self._connection_enabled = True
         self._last_message_time = time.monotonic()
         self._schedule_monitor()
-        # Trigger update of all attributes
+
+        await self._async_trigger_updates()
+
+    async def _async_trigger_updates(self) -> None:
+        """Trigger update of all attributes."""
         commands = [
-            "ZM?",
-            "SI?",
-            "MV?",
-            "MU?",
-            "Z2?",
-            "Z2MU?",
-            "Z3?",
-            "Z3MU?",
-            "PSTONE CTRL ?",
-            "PSBAS ?",
-            "PSTRE ?",
-            "PSDYNEQ ?",
-            "PSLFC ?",
-            "PSCNTAMT ?",
-            "PSMULTEQ: ?",
-            "PSREFLEV ?",
-            "PSDYNVOL ?",
-            "MS?",
-            "MNMEN?",
-            "DIM ?",
-            "PSDELAY?",
-            "ECO?",
-            "VSMONI ?",
-            "PSDIRAC ?",
-            "CV?",
-            "PSNEURAL ?",
-            "PSIMAX ?",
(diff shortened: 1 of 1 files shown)


henrikwidlund:
Nice, that seems like a good idea. I was thinking of using lower timeouts for the "important" calls, and then just the regular delay for the rest. Maybe you want to implement something similar to that? In my testing I made tuples of everything instead of just a set of strings where the second type was the delay.

ol-iver:
There are a couple of calls to which my receiver does not send a response, so it would take a lot more time until the update commands have all been sent with the default timeout. That's the main reason why I added the custom timeout. I don't think we are able to figure out which calls are valid for which receiver, so reducing the timeout is a mitigation for this.
The most important calls do return something and also do it quite fast.

ol-iver:
I just checked the logs. My receiver does not return anything to these 19 commands:
`PSNEURAL ?`, `PSIMAXAUD ?`, `PSIMAXSWM ?`, `SSTTR ?`, `BTTX ?`, `PSSPV ?`, `PSHEQ ?`, `PSDIRAC ?`, `PSIMAX ?`, `PSIMAXHPF ?`, `PSIMAXLPF ?`, `PSIMAXSWO ?`, `PSBSC ?`, `PSDEH ?`, `PSAUROPR ?`, `PSAUROST ?`, ` PSAUROMODE ?`, `SPPR ?`, `PSDIC ?` 

With the default timeout the update commands would run 34.2 seconds longer.

ol-iver:
Let's go ahead with this. We could also fine tune settings later if necessary 🙂

henrikwidlund:
@ol-iver I just realized something, using the timeout and waiting for response might not do what exactly we want, the receiver will not acknowledge the requests, so the library will always wait for .2 seconds, during which time nothing else will be able to be sent to the receiver. I guess it's possible for some other command to get sent between the iterations in the for loop for the registrations, but it'll be quite unpredictable.

I might be wrong, I'm not that proficient in Python.

ol-iver:
This depends primarly on your implementation.
The send command is protected by a lock and the startup commands try to acquire a new lock when the previous command has been completed.
The lock acquiring in python is fair, so the one who tries to aquire the lock first, will get it first.
If you would like to send multiple volume up commands and you implement it in a way that you always wait for one command to finish before you send the next one, there will be the sequence `startup`, `volume`, `startup`, `volume`, `startup`, `volume` 
If you send the commands all at the same time, they will all try to acquire the lock while the first startup command is running and it will result in a sequence like `startup`, `volume`, `volume`, `volume`, `startup`, `startup`

PR #353, opened 2025-10-28
Use httpx `stream` function instead of `get` and `post`

### LINKED ISSUE #352: discover() timeout when /goform/Deviceinfo.xml
```
Oct 28 19:44:03 assistant hass[1775988]: 2025-10-28 19:44:03.922 DEBUG (MainThread) [denonavr.decorators] HTTP timeout exception on request <Request('GET', 'http://10.250.7.14/goform/Deviceinfo.xml')>: 
Oct 28 19:44:03 assistant hass[1775988]: 2025-10-28 19:44:03.922 DEBUG (MainThread) [denonavr.foundation] Connection error on port 80 when identifying receiver: TimeoutException: 
```

This is caused by the receiver (specifically Marantz AV7702mkII) never closing the connection after returning 404 Not Found:

```
curl -v http://10.250.7.14:8080/goform/Deviceinfo.xml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 10.250.7.14:8080...
* Connected to 10.250.7.14 (10.250.7.14) port 8080
> GET /goform/Deviceinfo.xml HTTP/1.1
> Host: 10.250.7.14:8080
> User-Agent: curl/8.9.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 404 Not Found
* no chunk, no close, no size. Assume close to signal end
< 
  0     0    0     0    0     0      0      0 --:--:--  0:00:16 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:20 --:--:--     0{ [0 bytes data]
  0     0    0     0    0     0      0      0 --:--:--  0:00:20 --:--:--     0
* shutting down connection #0
```

It just stays there until timeout.

I assume since multiple requests are made that several paths are probed, but these paths should return immediately with error instead of waiting for the receiver to close the connection on its side.  It's probably necessary to tweak the headers being sent, or instead of waiting for the whole request to finish, call requests.raise_for_status() before reading the body, and if the status isn't 200 OK (or whatever was being expected), then unilaterally close the connection from the client side (this library).

Thanks for your help fixing this issue and helping many Home Assistant users in the process.

ol-iver:
This prevents that we wait until the connection times out when the receiver returns an HTTP error status but does not close the connection.

Fixes #352

### DIFF
--- a/denonavr/api.py
+++ b/denonavr/api.py
@@ -100,10 +100,11 @@ class HTTPXAsyncClient:
         """Call GET endpoint of Denon AVR receiver asynchronously."""
         client = self.client_getter()
         try:
-            res = await client.get(
-                url, timeout=httpx.Timeout(timeout, read=read_timeout)
-            )
-            res.raise_for_status()
+            async with client.stream(
+                "GET", url, timeout=httpx.Timeout(timeout, read=read_timeout)
+            ) as res:
+                res.raise_for_status()
+                await res.aread()
         finally:
             # Close the default AsyncClient but keep custom clients open
             if self.is_default_async_client():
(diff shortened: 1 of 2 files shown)


Rudd-O (approved):
LGTM but have you checked this works with a purposefully-broken HTTP server?  For all I know this also hangs because the server never finishes sending data to the client or closing the connection.

ol-iver:
When an exception is raised the context manager is exited. This closes the stream.
I tested it on a random non-existing endpoint of my receiver which does not close the connection too.

PR #356, opened 2025-10-31
Add missing sound modes

### LINKED ISSUE #355: Not able to match sound mode: 'DOLBY PLII MV', assuming 'DOLBY DIGITAL'.
I’m using a Denon AVR-S660H and occasionally see this message in the Home Assistant log:

`Not able to match sound mode: 'DOLBY PLII MV', assuming 'DOLBY DIGITAL'.`

Could you please add 'DOLBY PLII MV' under Movie in: https://github.com/ol-iver/denonavr/blob/main/denonavr/const.py (or somewhere else?)

While looking into this, I also found a recent documentation that might be useful: 
https://github.com/home-assistant/home-assistant.io/issues/38352

Thanks for your work!

henrikwidlund:
Also removes duplicates and some restructuring.

Fixes #355


denonavr/const.py
```diff
@@ -439,29 +490,60 @@
...
     ],
     "AURO3D": ["AURO-3D", "AURO3D"],
     "AURO2DSURR": ["AURO-2D SURROUND", "AURO2DSURR"],
     "MCH STEREO": [
         "MCH STEREO",
-        "MULTI CH IN 7.1",
         "MULTI CH STEREO",
         "MULTI_CH_STEREO",
     ],
+    "MCH": [
```
ol-iver: Do you have a `MCH` sound mode in your receiver? I don't.
The key must match a sound mode which exists on the receiver with this name.
Maybe the sound mode exist, I just would like to make you aware of this dependency.

denonavr/const.py
```diff
@@ -439,29 +490,60 @@
...
+        "MULTI CH IN 7.1",
+        "M CH IN+DOLBY EX",
+        "M CH IN+PL2Z H",
+        "M CH IN+DSUR",
+        "M CH IN+NEURAL:X",
+        "M CH IN+VIRTUAL:X",
+    ],
     "STEREO": ["STEREO"],
     ALL_ZONE_STEREO: ["ALL ZONE STEREO"],
+    "MPEG": [
```
ol-iver: Same for `MPEG`.

denonavr/const.py
```diff
@@ -439,29 +490,60 @@
...
+        "MPEG2 AAC",
+        "MPEG4 AAC",
+        "MPEG-H",
+    ],
+    "AAC": [
```
ol-iver: And for `AAC`.

ol-iver (commented):
Thanks 🙂

denonavr/const.py
henrikwidlund: Oh, I did not know about that dependency. The mpeg and aac are only available in Japan, so I can't verify that. But they all show up under both DD and DTS in the specification, same with the MCH section I added. Should I just put everything under DD, or can we have mappings in both of them?

henrikwidlund: I corrected it, but I don't understand how the MUSIC, MOVIE and GAME categories work, I've never seen those displayed on my unit, and all of the values in those dictionaries maps to DD or DTS in the Denon specification.

denonavr/const.py
```diff
@@ -415,6 +460,24 @@
...
+        "DOLBY D+ +PL2Z H",
+        "DOLBY HD+EX",
+        "DOLBY HD+PL2Z H",
+        "MPEG2 AAC",
+        "MPEG4 AAC",
+        "MPEG-H",
+        "AAC+DOLBY EX",
+        "AAC+PL2Z H",
+        "AAC+DSUR",
+        "MULTI CH IN",  # Exists in both DOLBY and DTS, but can't be in both
```
ol-iver: I don't understand the comment. This mode was only in `DTS SURROUND` before, now it is moved to `DOLBY DIGITAL`.
If both are working, then I prefer to keep it where it was.

denonavr/const.py
```diff
@@ -415,6 +460,24 @@
...
+        "MULTI CH IN 7.1",  # Exists in both DOLBY and DTS, but can't be in both
```
ol-iver: to keep it consistent with the previous comment, I'd rather move this to DTS too.

ol-iver (commented):
Before, the dicts were sorted in alphabetical order. Please sort them again this way.

denonavr/const.py
ol-iver: I don't know it either 😅 
Afaik MUSIC, MOVIE and GAME are buttons on the remote control of the receivers.
Maybe @starkillerOG could explain. He implemented sounds modes in the first place.

denonavr/const.py
henrikwidlund: I clarified it a bit with new comments. The spec from Denon says that these two modes exists for both DD and DTS, that's all.

denonavr/const.py
henrikwidlund: Haha, okay. I wonder if the "H" suffix stands for "SOURCE" somehow, see `DOLBY D+ +PL2Z H` for example. Not sure though, so I'll just leave them in DD and DTS.

starkillerOG: Yes the MUSIC, MOVIE, GAME are buttons on the remote.
They are also commands that can be sent to change to that mode on the receiver.

Unfortunately there are multiple modes that all fall under the same command, like Dolby pro logic music, or Dolby Neo music. On the remote you would repeatedly press the "Music" button to switch between those.

There is no command to directly set the receiver to dolby pro logic music for instance. But if the last music mode set was pro logic, then sending MUSIC command once will switch to that.

henrikwidlund: According to the manual, these are the different types of sound modes. They don't correspond exactly to how it's set up in the library.

https://manuals.denon.com/avca10h/eu/en/DRDZSYyrtgycpw.php

ol-iver: That's a user's manual. The interfaces might respond slightly different 😉 
You can have a look at the IP protocol specification: https://assets.denon.com/documentmaster/de/steuerungsprotokoll_ip_rs232c_avr-x1200w_avr-x2200w_avr-x3200w_avr-x4200w.pdf
Starting on page 4 you see the sound mode (`MS`) commands and the potential responses of a receiver of that type.

henrikwidlund: Yeah, that is similar to the spec I got from them for my model, but I was hoping to find some info on whether or not the Japanese specific ones would be in DD/DTS or its own section. I think the sorting and grouping now follows the rationale you described.

ol-iver: The receiver interfaces of different models are more complicated then they should be. Different ports here and there, slightly different naming. It takes much time to figure all these things out.

ol-iver (approved):
Thanks for addressing my feedback 🙂

PR #357, opened 2025-11-03
Misc code fixes and cleanups

henrikwidlund:
- Moves fields without defaults to the top to make PyCharm happy
- Removes redundant cast
- Removes redundant int parsing from string
- Optimize regex
- Fixes return types

### DIFF
--- a/denonavr/api.py
+++ b/denonavr/api.py
@@ -465,6 +465,8 @@ class DenonAVRTelnetProtocol(asyncio.Protocol):
 class DenonAVRTelnetApi:
     """Handle Telnet responses from the Denon AVR Telnet interface."""
 
+    _send_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
+    _update_callback_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
     host: str = attr.ib(converter=str, default="localhost")
     timeout: float = attr.ib(converter=float, default=2.0)
     is_denon: bool = attr.ib(converter=bool, default=True)
@@ -483,7 +485,6 @@ class DenonAVRTelnetApi:
         default=attr.Factory(asyncio.Event)
     )
     _send_confirmation_command: str = attr.ib(converter=str, default="")
-    _send_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
     _callbacks: Dict[str, List[Callable]] = attr.ib(
         validator=attr.validators.instance_of(dict),
         default=attr.Factory(dict),
@@ -494,7 +495,6 @@ class DenonAVRTelnetApi:
         default=attr.Factory(list),
         init=False,
     )
-    _update_callback_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
 
     def __attrs_post_init__(self) -> None:
         """Initialize special attributes."""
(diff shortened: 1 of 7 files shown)


henrikwidlund:
@ol-iver I added `# TODO: This method doesn't seem to exist in DenonAVRFoundation?`, I don't understand how the `async_setup` for other zones can work as it is today since the type doesn't include that method

denonavr/api.py
```diff
@@ -465,6 +465,8 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
 class DenonAVRTelnetApi:
     """Handle Telnet responses from the Denon AVR Telnet interface."""
 
+    _send_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
+    _update_callback_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
```
ol-iver: Why putting these in front?
I prefer having public fields first and then private ones. Please keep the order.

This is probably the way to go.
```python
    _send_tasks: Set[asyncio.Task] = attr.ib(default=attr.Factory(set))
...
    _update_callback_tasks: Set[asyncio.Task] = attr.ib(default=attr.Factory(set))
```

denonavr/input.py
```diff
@@ -94,6 +94,7 @@ def set_input_func(
 class DenonAVRInput(DenonAVRFoundation):
     """This class implements input functions of Denon AVR receiver."""
 
+    _callback_tasks: Set[asyncio.Task] = attr.ib(attr.Factory(set))
```
ol-iver: Similar to a previous change
```python
    _callback_tasks: Set[asyncio.Task] = attr.ib(default=attr.Factory(set))
```
at its original place.

denonavr/input.py
```diff
@@ -532,7 +532,7 @@ async def async_get_changed_sources_appcommand(
 
         self._replace_duplicate_sources(renamed_sources)
 
-        return (renamed_sources, deleted_sources)
+        return renamed_sources, deleted_sources
```
ol-iver, line 535: The next method returns a tuple too. Please change it accordingly.

denonavr/denonavr.py
```diff
@@ -175,7 +175,7 @@ async def async_setup(self) -> None:
 
             for zone_name, zone_item in self._zones.items():
                 if zone_name != self.zone:
-                    await zone_item.async_setup()
+                    await zone_item.async_setup()  # TODO: This method doesn't seem to exist in DenonAVRFoundation?
```
ol-iver: The zone is of type `DenonAVR` too (see[ here](https://github.com/ol-iver/denonavr/blob/680c33d5347c50660bd82603e5568dd1df75bbce/denonavr/denonavr.py#L150-L157)). That's why it is working 😄 

However, I used the parent class (`DenonAVRFoundation`) as type because the class cannot reference itself in type annotations.
You could try this in line 88 of the same file.
```python
    _zones: Dict[str, "DenonAVR"] = attr.ib(
```

ol-iver (changes requested):
Thanks for improving the code quality 😄

denonavr/api.py
henrikwidlund: It was because PyCharm complains. I'll not your preference for future PRs (I'm the other way around). I've moved them back to their original rows and added defaults. Thanks for the feedback 🙂

ol-iver (approved):
🚀

PR #360, opened 2025-11-15
Add additional commands

henrikwidlund:
- Dolby Atmos Toggle
- Page/Ch Up/Down
- Input Mode (Select, Auto, HDMI, Digital, Analog)


ol-iver (approved):
Thanks for adding these new commands 🙂

Issue #335, opened 2025-05-14
M-CR510


M4TH1EU:
I keep getting these errors from denonavr running latest homeassistant and a Marantz M-CR510:

```
2025-05-14 18:04:09.291 INFO (MainThread) [denonavr.input] No mapping for source Media Server found
2025-05-14 18:01:19.273 INFO (MainThread) [denonavr.input] No mapping for source NET found
```

I've ran the script you asked people with the issues to run, here are the various XML files it created:
[AVR-M-CR510.zip](https://github.com/user-attachments/files/20210993/AVR-M-CR510.zip)

ol-iver:
It looks like your receiver returns the same device sometimes as `Media Server` and sometimes as `Music Server`.

Are you able to change the input to Media/Music Server using homeassistant?

PR #362, opened 2025-12-11
Minor fixes and add tests for Denon CEOL piccolo DRA-N4

baracudaz:
* Fix issue with caching throwing [exceptions](https://github.com/user-attachments/files/24099747/HA.txt) in Home Assistant 
* Add missing input to allow Internet Radio and Media Server inputs to be displayed correctly in Home Assistant:
```
2025-12-10 15:38:42.436 INFO (MainThread) [denonavr.input] No mapping for source NET found
2025-12-10 16:43:48.207 INFO (MainThread) [denonavr.input] No mapping for source Media Server found
```
Relates to https://github.com/ol-iver/denonavr/issues/335

* Add tests and mocks for [Denon CEOL piccolo DRA-N4](https://assets.denon.com/documentmaster/uk/dn_ceol_piccolo_productinfo_pdf_v3_en_28102014.pdf)
* Fixes for typos and URLs
 
TODO: Attribute names command_cusor_* (misspelled) should be cursor


denonavr/api.py
```diff
@@ -112,7 +112,6 @@ async def async_get(
 
         return res
 
-    @cache_result
```
ol-iver, line 115: The cache should not be removed entirely, otherwise the `async_update` would take much longer.

Instead you could fix the cache. [This part](https://github.com/ol-iver/denonavr/blob/07238f45fa56fda9f6a447270dfa68f8e3442f8a/denonavr/decorators.py#L79-L80) could be replaced by something like this
```
        if kwargs.get("cache_id") is None:
            return await func(*args, **kwargs)
```
This way, the `lru_cache` would not be used if the command should be cached.

ol-iver (changes requested):
Thanks for fixing these issues. I have one suggestion though.

ol-iver (approved):
Nice 🚀

denonavr/api.py
baracudaz: Sure, great suggestion.

PR #369, opened 2026-02-01
Fix some telnet commands

ol-iver:
Some telnet commands which have been added in #360 look wrong.

@henrikwidlund please have a look `/goform/formiPhoneAppDirect.xml?RCKSK0410348` etc. do not look like correct telnet commands.


henrikwidlund:
@ol-iver if the telnet commands have goform stuff in them, then that is wrong, it should only be the IR code identifier.

ol-iver:
Yeah, that was the case. This is fixed now. Thanks for your feedback 🙂

PR #380, opened 2026-02-07
Fix update_callback example in README

mvdwetering:
The example in the README with the update_callback has async in front of it which causes it to fail. Removing async makes it work fine.


ol-iver (approved):
Thanks 🙂

PR #381, opened 2026-02-12
Add hdmiout on http

mvdwetering:
This PR add support for HDMI out using the xml api with the goal to later make it available in Home Assistant (it is my current understanding that the XML suppport is needed for that).

I tried to build it in a similar way to how other things are done. I used the naming as used in the XML which is videoout iso hdmiout, but kept the current API the same.

Manual testing with a simple example script seems to indicate it works with my receiver (AVC-X2850H).

However...

Some tests started failing and I don't really understand why.
Also I could not  figure out how these tests work exactly. They seem to use the XML files that are in the `tests/xml` directory, but unclear to me how to update those files and if that should be even needed.

I would appreciate some pointers on how to approach the testing part.

### DIFF
--- a/denonavr/appcommand.py
+++ b/denonavr/appcommand.py
@@ -145,6 +145,27 @@ class AppCommands:
         ),
     )
 
+    GetOutputSettings = AppCommandCmd(
+        cmd_id="3",
+        name="GetOutputSettings",
+        param_list=(
+            AppCommandCmdParam(name="videoout"),
+        ),
+        response_pattern=(
+            AppCommandResponsePattern(
+                update_attribute="_videoout",
+                add_zone=False,
+                suffix="/list/param[@name='videoout']",
+            ),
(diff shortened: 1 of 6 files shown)


ol-iver:
The problem is that your new commands are using the `/goform/AppCommand0300.xml` interface and there is no mock interface implemented like it is for other endpoints e.g. for `/goform/Appcommand.xml` (see [here](https://github.com/ol-iver/denonavr/blob/ff18d60468722c50e48f9c1bcc97da4f2f841d51/tests/test_denonavr.py#L127)).
Thus, the mock interface always return `DATA` (see [here](https://github.com/ol-iver/denonavr/blob/ff18d60468722c50e48f9c1bcc97da4f2f841d51/tests/test_denonavr.py#L142-L143)), which is the reason for the `XMLParseError`s.

Issue #375, opened 2026-02-04
Sync of changes


henrikwidlund:
Hi @ol-iver, as you know I've forked this repo and I've done some modifications. I'm wondering if you want me to contribute some of them to your repo? You can find the diff here: https://github.com/ol-iver/denonavr/compare/main...henrikwidlund:denonavr:main?expand=1

List of things I've added:
- mDNS discovery (my unit and others who are using the UC remote no longer functions with SSDP).
  -  This addition also filters on device type so that discovered devices via mDNS and SSDP only contains those who are receivers.
- Rate limiting
  - The receivers usually don't like being spammed with requests which would happen in our case if you held down the volume button for example.
    - The rate limiter is dynamic and takes response times into consideration.
- Custom event for max volume
  - The library used to ignore the MV event when parameter was MAX
  - We started using this because when the max allowed volume is reached the receiver would stop responding when you tried to further increase the volume causing timeouts
- Check current value for a command
  - Sending the same value or a value that is already at min or max causes the receiver to not send a response, timeout again
  - Tests to verify all of this
- Added delay command to set a specific delay
- Added mute toggle command
- Corrected some docs in regards to which values that are accepted
- Changed IMAX filters to use integers instead of strings
- Parallelize some calls made within the setup and update methods
- Added properties for the following:
  - Audio Signal
  - Audio Sound
  - Audio Sampling Rate
  - Video Signal In/Out
  - Input/Output Channels (7.1.4 for example)
  - HDR Input/Output Info
  - Pixel Depth Input/Output
  - Mx FRL Input/Output (mostly seem to be empty, but that is also what the receiver itself shows, maybe the feature isn't completed)
  - Colorspace Input/Output
  - Max Resolution (4K18Gbps for example)
  - The properties above related to video require polling over telnet, they are not propagated reliably even if subscribed to)
- Adjusted the code that handles volume parsing as we noticed that the volume could go out of range (not sure if this was a bug I introduced in my for or not)

dmutsaers:
I'm looking forward to see these changes added to `denonavr`.

ol-iver:
That would be nice @henrikwidlund 🙂 
Could you open own PRs for each item? This would be easier to review. I lost track when I just went through all commits 😅

henrikwidlund:
Okay @ol-iver, I've started the work of adding all of it, starting with the rate limiter. Once that is merged I'll hopefully be able to do a few in parallel without introducing conflicts.

PR #383, opened 2026-04-19
Add rate limiter

henrikwidlund:
First of many PRs to come :)

Relates to #375

### DIFF
--- a/denonavr/api.py
+++ b/denonavr/api.py
@@ -47,6 +47,7 @@ from .exceptions import (
     AvrProcessingError,
     AvrTimoutError,
 )
+from .rate_limiter import AdaptiveLimiter
 
 if sys.version_info[:2] < (3, 11):
     from async_timeout import timeout as asyncio_timeout
@@ -83,6 +84,11 @@ class HTTPXAsyncClient:
         init=False,
     )
 
+    rate_limiter: AdaptiveLimiter = attr.ib(
+        validator=attr.validators.instance_of(AdaptiveLimiter),
+        default=attr.Factory(AdaptiveLimiter),
+    )
+
     def __hash__(self) -> int:
         """Hash the class using its ID that caching works."""
         return id(self)
@@ -92,14 +98,20 @@ class HTTPXAsyncClient:
     async def async_get(
         self,
         url: str,
+        rate_limit_key: str,
         timeout: float,
         read_timeout: float,
         *,
         cache_id: Hashable = None,
+        record_latency: bool = True,
+        skip_rate_limiter: bool = False,
     ) -> httpx.Response:
         """Call GET endpoint of Denon AVR receiver asynchronously."""
         client = self.client_getter()
+        start = time.monotonic()
         try:
+            if not skip_rate_limiter:
+                await self.rate_limiter.acquire(rate_limit_key)
             async with client.stream(
                 "GET", url, timeout=httpx.Timeout(timeout, read=read_timeout)
             ) as res:
@@ -110,6 +122,8 @@ class HTTPXAsyncClient:
             if self.is_default_async_client():
                 await client.aclose()
 
+        if record_latency:
+            self.rate_limiter.record_latency(rate_limit_key, start)
         return res
 
     @cache_result
@@ -117,16 +131,22 @@ class HTTPXAsyncClient:
     async def async_post(
         self,
         url: str,
+        rate_limit_key: str,
         timeout: float,
         read_timeout: float,
         *,
         content: Optional[bytes] = None,
         data: Optional[Dict] = None,
         cache_id: Hashable = None,
+        record_latency: bool = True,
+        skip_rate_limiter: bool = False,
     ) -> httpx.Response:
         """Call POST endpoint of Denon AVR receiver asynchronously."""
         client = self.client_getter()
+        start = time.monotonic()
         try:
+            if not skip_rate_limiter:
+                await self.rate_limiter.acquire(rate_limit_key)
             async with client.stream(
                 "POST",
                 url,
@@ -141,11 +161,17 @@ class HTTPXAsyncClient:
             if self.is_default_async_client():
                 await client.aclose()
 
+        if record_latency:
+            self.rate_limiter.record_latency(rate_limit_key, start)
         return res
 
     def is_default_async_client(self) -> bool:
         """Check if default httpx.AsyncClient getter is used."""
         return self.client_getter is get_default_async_client
+
+    async def aclose(self):
+        """Close rate limiter when done."""
+        await self.rate_limiter.aclose()
 
 
 @attr.s(auto_attribs=True, on_setattr=DENON_ATTR_SETATTR)
@@ -175,6 +201,7 @@ class DenonAVRApi:
         default=attr.Factory(HTTPXAsyncClient),
         init=False,
     )
+    _http_callback_tasks: Set[asyncio.Task] = attr.ib(default=attr.Factory(set))
 
     async def async_get(
         self,
@@ -182,6 +209,9 @@ class DenonAVRApi:
         *,
         port: Optional[int] = None,
         cache_id: Hashable = None,
+        skip_confirmation: bool = False,
+        record_latency: bool = True,
+        skip_rate_limiter: bool = False,
     ) -> httpx.Response:
         """Call GET endpoint of Denon AVR receiver asynchronously."""
         # Use default port of the receiver if no different port is specified
@@ -189,8 +219,31 @@ class DenonAVRApi:
 
         endpoint = f"http://{self.host}:{port}{request}"
 
+        if skip_confirmation:
+            task = asyncio.create_task(
+                self.httpx_async_client.async_get(
+                    endpoint,
+                    self.host,
+                    self.timeout,
+                    self.read_timeout,
+                    cache_id=cache_id,
+                    record_latency=False,
+                    skip_rate_limiter=skip_rate_limiter,
+                )
+            )
+            task.add_done_callback(
+                self._http_callback_tasks.discard
+            )  # Prevent garbage collection
+            return httpx.Response(200, text="")
+
         return await self.httpx_async_client.async_get(
-            endpoint, self.timeout, self.read_timeout, cache_id=cache_id
+            endpoint,
+            self.host,
+            self.timeout,
+            self.read_timeout,
+            cache_id=cache_id,
+            record_latency=record_latency,
+            skip_rate_limiter=skip_rate_limiter,
         )
 
     async def async_post(
@@ -210,17 +263,29 @@ class DenonAVRApi:
 
         return await self.httpx_async_client.async_post(
             endpoint,
+            self.host,
             self.timeout,
             self.read_timeout,
             content=content,
             data=data,
             cache_id=cache_id,
+            record_latency=False,
+            skip_rate_limiter=True,
         )
 
-    async def async_get_command(self, request: str) -> str:
+    async def async_get_command(
+        self,
+        request: str,
+        skip_confirmation: bool = False,
+        skip_rate_limiter: bool = False,
+    ) -> str:
         """Send HTTP GET command to Denon AVR receiver asynchronously."""
         # HTTP GET to endpoint
-        res = await self.async_get(request)
+        res = await self.async_get(
+            request,
+            skip_confirmation=skip_confirmation,
+            skip_rate_limiter=skip_rate_limiter,
+        )
         # Return text
         return res.text
 
@@ -229,7 +294,9 @@ class DenonAVRApi:
     ) -> ET.Element:
         """Return XML data from HTTP GET endpoint asynchronously."""
         # HTTP GET to endpoint
-        res = await self.async_get(request, cache_id=cache_id)
+        res = await self.async_get(
+            request, cache_id=cache_id, record_latency=False, skip_rate_limiter=True
+        )
         # create ElementTree
         try:
             xml_root = fromstring(res.text)
@@ -494,6 +561,11 @@ class DenonAVRTelnetApi:
         default=attr.Factory(list),
         init=False,
     )
+    _rate_limiter: AdaptiveLimiter = attr.ib(
+        validator=attr.validators.instance_of(AdaptiveLimiter),
+        default=attr.Factory(AdaptiveLimiter),
+        init=False,
+    )
     _update_callback_tasks: Set[asyncio.Task] = attr.ib(default=attr.Factory(set))
 
     def __attrs_post_init__(self) -> None:
@@ -635,6 +707,8 @@ class DenonAVRTelnetApi:
         await self.async_send_commands(
             *commands,
             confirmation_timeout=0.2,
+            skip_rate_limiter=True,
+            record_latency=False,
         )
 
     def _schedule_monitor(self) -> None:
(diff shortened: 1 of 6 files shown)


denonavr/api.py
```diff
@@ -175,22 +201,49 @@ class DenonAVRApi:
...
+                    cache_id=cache_id,
+                    record_latency=False,
+                    skip_rate_limiter=skip_rate_limiter,
+                )
+            )
+            task.add_done_callback(
+                self._http_callback_tasks.discard
+            )  # Prevent garbage collection
+            return httpx.Response(200, text="")
+
```
ol-iver: Creating a task for each HTTP request is too much compute overhead and faking the response is no good practice. Please remove the `skip_confirmation` part for HTTP requests.

denonavr/rate_limiter.py
```diff
@@ -0,0 +1,167 @@
...
+      based on the EWMA of observed latencies (excluding skip_confirmation calls).
+    - Burst protection is inherent in the token bucket model.
+    """
+
+    def __init__(
+        self,
+        *,
+        initial_wait_ms: float = 100.0,
+        min_wait_ms: float = 100.0,
+        max_wait_ms: float = 200.0,
```
ol-iver: Why do you call this `..._wait_ms`? This gives the impression that even the first request needs to wait 100ms.
Later you even calculate rates from this wait times. Please call it `rate` from the beginning.

Also the minimum rate (10 req/s) looks too low. It could be 100 req/s.

denonavr/api.py
```diff
@@ -898,20 +978,26 @@ async def _async_send_command(
                         "Timeout waiting for confirmation of command: %s", command
                     )
                 finally:
+                    if record_latency:
+                        self._rate_limiter.record_latency(self.host, start)
```
ol-iver: You could move this from `finally` to `else`.
The default timeout is 2s. When the receiver does not send a confirmation within this timeframe, the command might never confirm. This should not affect the rate limit.

denonavr/api.py
```diff
@@ -141,12 +161,18 @@ async def async_post(
...
+            self.rate_limiter.record_latency(rate_limit_key, start)
         return res
 
     def is_default_async_client(self) -> bool:
         """Check if default httpx.AsyncClient getter is used."""
         return self.client_getter is get_default_async_client
 
+    async def aclose(self):
+        """Close rate limiter when done."""
+        await self.rate_limiter.aclose()
```
ol-iver: This method isn't called anywhere.

denonavr/rate_limiter.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Safe to call even if key doesn't exist yet; if the key is not present,
+        the call is silently ignored and no action is taken.
+        """
+        seconds = time.monotonic() - start
+        # Only record if destination already initialized
+        if latency_tracker := self._latencies.get(destination):
+            latency_tracker.update(seconds)
+
+    async def aclose(self) -> None:
```
ol-iver: The only place where this method is called, is never called 😉
There must be a better way to do this.

denonavr/rate_limiter.py
```diff
@@ -0,0 +1,167 @@
...
+            while not self._shutdown_event.is_set():
+                await asyncio.sleep(self._adjust_interval)
+                if self._shutdown_event.is_set():
+                    break
+                avg = self._latencies[key].value
+                new_rate = self._target_rate(avg)
+                # AsyncLimiter does not expose direct rate mutation;
+                # recreate it atomically under a lock to avoid races.
+                async with self._locks[key]:
+                    self._limiters[key] = AsyncLimiter(new_rate, time_period=1)
```
ol-iver: Why do you recreate the limiter every 2 seconds even when there are no changes? This does not look efficient. 

You could start a task adjust the rate limiter in `record_latency` if the current avg rate and the limiter rate differ by more than 20% and maximum once every 10s.
Then you can also remove the update loop which cannot be stopped.

denonavr/api.py
```diff
@@ -635,6 +707,8 @@ async def _async_trigger_updates(self) -> None:
         await self.async_send_commands(
             *commands,
             confirmation_timeout=0.2,
+            skip_rate_limiter=True,
+            record_latency=False,
```
ol-iver: Many commands hammer on the receiver in a short period of time. Isn't this the perfect place to use rate limiting?

denonavr/rate_limiter.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Record RTT after a call.
+
+        Safe to call even if key doesn't exist yet; if the key is not present,
+        the call is silently ignored and no action is taken.
+        """
+        seconds = time.monotonic() - start
+        # Only record if destination already initialized
+        if latency_tracker := self._latencies.get(destination):
+            latency_tracker.update(seconds)
```
ol-iver: Could you skip the update for big values like >1s. 
The audyssey endpoint of my receiver take like 4 seconds to respond and I don't want the rate limit go down just because I call it a couple of times.

denonavr/api.py
```diff
@@ -886,6 +963,9 @@ async def _async_send_command(
                     f"Error sending command {command}. Telnet connected: "
                     f"{self.connected}, Connection healthy: {self.healthy}"
                 )
+            start = time.monotonic()
+            if not skip_rate_limiter:
+                await self._rate_limiter.acquire(self.host)
```
ol-iver: Please use a different limiter for telnet as HTTP and telnet are also different endpoints.

denonavr/api.py
```diff
@@ -210,17 +263,29 @@ async def async_post(
...
         return await self.httpx_async_client.async_post(
             endpoint,
+            self.host,
             self.timeout,
             self.read_timeout,
             content=content,
             data=data,
             cache_id=cache_id,
+            record_latency=False,
+            skip_rate_limiter=True,
```
ol-iver: Why is this method not rate limited?
In my opinion all requests should go though the rate limiter if it is enabled.

ol-iver (changes requested):
Thanks for the rate limiter. It's nice 😄

Please add a switch to enable/disable the rate limiter for a receiver instance in general.

denonavr/api.py
henrikwidlund: Fair enough, I don't think it's actually needed anyway. From what I remember it was a "red herring", and the delay I saw came from something else.

denonavr/rate_limiter.py
henrikwidlund: I landed on 10 rps by sending volume up/down commands with different delays in-between, and landed on 10 because that seemed like a good balance with speed for the command without causing the receiver to buffer the commands and then keep executing them for seconds after I stopped sending commands. My thought was that it is more irritating to not be able to send a different command after you let go of a button than having volume not increase by more than 10 steps per second. I'm open for adjusting the 10rps, but 100 will cause issues.

denonavr/api.py
henrikwidlund: Yes, I think this one slipped through, it is called in my fork in code that has not been included here. Do you want me to delete it or replace it with something else so that the resource is disposed when no longer needed?

denonavr/api.py
henrikwidlund: Fixed, also found another place where it applies

denonavr/api.py
henrikwidlund: This is yet another case of mismatch between this repo and my fork, in my work I have a delay between each command so the rate limiter is not needed there and I didn't want the automatic adjust of the rate limiter to be affected since I did do my own measurements to make sure it is as quick as possible without stalling the receiver. I will change it to work with your repo.

denonavr/rate_limiter.py
henrikwidlund: I think 1s might be too aggressive and might sidestep the intent of the limiter. I'll exclude those that take 2s. Let me know if you disagree and I can change it to 1.

denonavr/api.py
henrikwidlund: Also because of how the code is used in my fork, but you're right, in a general purpose library it makes sense to rate limit everything

denonavr/api.py
henrikwidlund: They are different, _rate_limiter is constructed by attr in the DenonAVRTelnetApi and HTTPXAsyncClient, or are you referring to using different settings?

denonavr/rate_limiter.py
henrikwidlund: Seems my comment disappeared, it is a leftover from my own fork that does things differently. Fixed it (I hope)

denonavr/rate_limiter.py
henrikwidlund: Adjusted

denonavr/foundation.py
```diff
@@ -1094,7 +1096,9 @@ async def async_cursor_up(self) -> None:
                 self.telnet_commands.command_cusor_up, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_cusor_up)
+            await self.api.async_get_command(
+                self.urls.command_cusor_up,
+            )
```
ol-iver: Please revert this to have a single line statement again.

denonavr/foundation.py
```diff
@@ -1103,7 +1107,9 @@ async def async_cursor_down(self) -> None:
                 self.telnet_commands.command_cusor_down, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_cusor_down)
+            await self.api.async_get_command(
+                self.urls.command_cusor_down,
+            )
```
ol-iver: Same here.

denonavr/foundation.py
```diff
@@ -1112,7 +1118,9 @@ async def async_cursor_left(self) -> None:
                 self.telnet_commands.command_cusor_left, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_cusor_left)
+            await self.api.async_get_command(
+                self.urls.command_cusor_left,
+            )
```
ol-iver: and here

denonavr/foundation.py
```diff
@@ -1121,7 +1129,9 @@ async def async_cursor_right(self) -> None:
                 self.telnet_commands.command_cusor_right, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_cusor_right)
+            await self.api.async_get_command(
+                self.urls.command_cusor_right,
+            )
```
ol-iver: and here

denonavr/foundation.py
```diff
@@ -1130,7 +1140,9 @@ async def async_cursor_enter(self) -> None:
                 self.telnet_commands.command_cusor_enter, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_cusor_enter)
+            await self.api.async_get_command(
+                self.urls.command_cusor_enter,
+            )
```
ol-iver: and here

denonavr/foundation.py
```diff
@@ -1139,7 +1151,9 @@ async def async_back(self) -> None:
                 self.telnet_commands.command_back, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_back)
+            await self.api.async_get_command(
+                self.urls.command_back,
+            )
```
ol-iver: and here

denonavr/foundation.py
```diff
@@ -1148,7 +1162,9 @@ async def async_info(self) -> None:
                 self.telnet_commands.command_info, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_info)
+            await self.api.async_get_command(
+                self.urls.command_info,
+            )
```
ol-iver: and here

denonavr/foundation.py
```diff
@@ -1157,7 +1173,9 @@ async def async_options(self) -> None:
                 self.telnet_commands.command_options, skip_confirmation=True
             )
         else:
-            await self.api.async_get_command(self.urls.command_options)
+            await self.api.async_get_command(
+                self.urls.command_options,
+            )
```
ol-iver: and here

denonavr/api.py
```diff
@@ -229,7 +269,9 @@ async def async_get_xml(
     ) -> ET.Element:
         """Return XML data from HTTP GET endpoint asynchronously."""
         # HTTP GET to endpoint
-        res = await self.async_get(request, cache_id=cache_id)
+        res = await self.async_get(
+            request, cache_id=cache_id, record_latency=False, skip_rate_limiter=True
```
ol-iver: This request should be rate limited too, shouldn't it?
You could remove the `skip_rater_limiter` option entirely, since it is not used anywhere else.

denonavr/rate_limiter.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class AdaptiveLimiter:
+    """
+    Adaptive token-bucket rate limiter per key.
+
+    Parameters:
+    - initial_rate: initial wait time between requests in milliseconds (> 0)
+    - min_rate: minimum wait time between requests in milliseconds (> 0)
+    - max_rate: maximum wait time between requests in milliseconds (>= min_rate)
```
ol-iver: Please update the descriptions. It should be rate instead of wait time now 🙂

denonavr/api.py
```diff
@@ -636,6 +683,8 @@ async def _async_trigger_updates(self) -> None:
         await self.async_send_commands(
             *commands,
             confirmation_timeout=0.2,
+            skip_rate_limiter=False,
+            record_latency=False,
```
ol-iver: The latency of those requests should be recorded too.
Then you could remove the `record_latency` option entirely.

ol-iver (changes requested):
Nice, thank you for keeping up and adressing my feedback 😄 

Please  remove the `skip_rate_limiter` and `record_latency` options completely and add an option to the receiver classes to enable/disable the rate limiter completely.

denonavr/api.py
ol-iver: I did not see this. All good then 🙂

denonavr/rate_limiter.py
ol-iver: Shouldn't the rate limiter find a proper value by itself? If 100 is too much for a receiver it should throttle the requests until it works 🙂

henrikwidlund: I guess we can leave the initial rate at 100ms and then let the time span for the request adjust the actual rate limit. I do however feel having a floor is useful, maybe just change it to 10ms or something to avoid real issues? I'm still a bit hesitant though, this change would result in a lot of commands being buffered, but to be fair, it just moves the problem from the API layer to the actual client.

denonavr/api.py
henrikwidlund: Fixed.

denonavr/rate_limiter.py
henrikwidlund: Fixed.

denonavr/api.py
henrikwidlund: Fixed.

denonavr/rate_limiter.py
henrikwidlund: I decided to restructure the rate limiter a bit, I think it's easier to reason about what the code does now. I also changed the initial RPS to 25, min rate 5 and max rate 100

ol-iver (approved):
Nice, thank you 🚀

PR #385, opened 2026-04-23
Fix one-zone telnet power refresh handling

davidjirovec:
## Summary
- use PW consistently for one-zone telnet power refreshes during reconnect
- accept PW events routed to the All zone for main-zone power/state callbacks
- add regression coverage for one-zone PW handling and reconnect ordering

## Problem
One-zone receivers can flap from off to on during telnet reconnect. The library already switched one-zone power control to PW, but telnet reconnect still queried ZM first and the main-zone power/state callbacks ignored PW events because they were routed to the All zone. That allowed stale power=ON state to survive long enough for an SI event to flip the media-player state back to on while the AVR was still in standby.

## Testing
- python3 -m compileall denonavr tests
- python3 -m pytest -q -s

### DIFF
--- a/denonavr/api.py
+++ b/denonavr/api.py
@@ -468,6 +468,7 @@ class DenonAVRTelnetApi:
     host: str = attr.ib(converter=str, default="localhost")
     timeout: float = attr.ib(converter=float, default=2.0)
     is_denon: bool = attr.ib(converter=bool, default=True)
+    power_query_command: str = attr.ib(converter=str, default="ZM?")
     _connection_enabled: bool = attr.ib(default=False)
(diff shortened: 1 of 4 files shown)


ol-iver (approved):
Nice, thanks for fixing this bug 🙂

ol-iver:
I fixed the import order.

davidjirovec:
Will there soon be release, so the fix will get to Home Assistant? Thanks :)

ol-iver:
> Will there soon be release, so the fix will get to Home Assistant? Thanks :)

done 🙂

PR #389, opened 2026-06-08
Include the request URL in HTTP error messages

### LINKED ISSUE #388: Include the request URL in HTTP error messages (`AvrNetworkError` & siblings)
## Problem

When a request fails at the connection level, `async_handle_receiver_exceptions`
in `denonavr/decorators.py` raises e.g.:

```python
raise AvrNetworkError(f"NetworkError: {err}", err.request) from err
```

The `httpx` error's `str()` is often just `"All connection attempts failed"`, so
the surfaced message becomes:

```
NetworkError: All connection attempts failed
```

The actual request URL — which pinpoints *what* couldn't be reached — is already captured in `err.request` (and even logged at debug level), but it never makes it into the exception message. `AvrRequestError` stores `self.request` but only forwards `message` to `Exception`, so `str(ex)` has no host/port.

Downstream consumers like the Home Assistant integration re-raise this as-is (`raise ConfigEntryNotReady from ex`), so the user sees only `NetworkError: All connection attempts failed` with nothing else.

## Real-world impact

A Marantz NR1508's UPnP description service (port `60006`) became unresponsive while the rest of the device stayed reachable (telnet on 23 and `Deviceinfo.xml` on 8080 both fine). HA setup failed with `NetworkError: All connection attempts failed` and nothing else — no hint that `:60006` specifically was the dead endpoint. The URL was right there in `err.request.url`.

## Suggested fix

Include `err.request.url` in the raised message across the handlers in `decorators.py`. The data is already on hand; it's a one-line-per-handler change.
After the fix the message reads e.g.:

```
NetworkError for http://192.168.178.123:60006/upnp/desc/aios_device/aios_device.xml: All connection attempts failed
```

(PR follows)

1358:
## Summary

Errors raised by `async_handle_receiver_exceptions` now include the failing request URL in their message. Previously only the bare transport error was surfaced, e.g.:

```
NetworkError: All connection attempts failed
```

…with no indication of *which* endpoint or port could not be reached. The URL was already captured in `err.request` (and logged at debug level), but `AvrRequestError` forwards only `message` to `Exception`, so `str(ex)` dropped it. Consumers such as the Home Assistant integration re-raise the error as-is (`raise ConfigEntryNotReady from ex`), so the user never saw the URL.

After this change the message reads e.g.:

```
NetworkError for http://192.168.0.50:60006/upnp/desc/aios_device/aios_device.xml: All connection attempts failed
```

## Motivation / context

While debugging a Marantz NR1508 that had dropped off Home Assistant, the only symptom was `NetworkError: All connection attempts failed`. The actual cause was the receiver's UPnP description service on port `60006` having become unresponsive (the rest of the device was reachable). Diagnosis took far longer than it should have purely because the failing URL — already present on the exception object — was not shown.

Closes #388.

## Changes

- `denonavr/decorators.py`: add `err.request.url` to the message of every error raised in `async_handle_receiver_exceptions` (`AvrForbiddenError`, `AvrRequestError`, `AvrTimoutError`, `AvrNetworkError`,   `AvrInvalidResponseError`). The `request` object stays attached unchanged.
- `tests/test_denonavr.py`: new parametrized test asserting the URL appears in `str(ex)` for the network-error and timeout paths, and that the request object is still exposed.

## Testing

```
$ pytest tests/
37 passed
```

## Notes

- No public API change: only message text is enriched; the second positional `request` argument and the `.request` attribute are unchanged.
- The pre-existing class-name typo `AvrTimoutError` is intentionally left as-is here to avoid an API-breaking rename in this PR.

### DIFF
--- a/denonavr/decorators.py
+++ b/denonavr/decorators.py
@@ -39,21 +39,29 @@ def async_handle_receiver_exceptions(func: Callable[..., AnyT]) -> Callable[...,
             _LOGGER.debug("HTTP status error on request %s: %s", err.request, err)
             # Separate handling of 403 errors
             if err.response.status_code == 403:
-                raise AvrForbiddenError(f"HTTPStatusError: {err}", err.request) from err
-            raise AvrRequestError(f"HTTPStatusError: {err}", err.request) from err
+                raise AvrForbiddenError(
+                    f"HTTPStatusError for {err.request.url}: {err}", err.request
+                ) from err
+            raise AvrRequestError(
+                f"HTTPStatusError for {err.request.url}: {err}", err.request
+            ) from err
         except httpx.TimeoutException as err:
             _LOGGER.debug("HTTP timeout exception on request %s: %s", err.request, err)
-            raise AvrTimoutError(f"TimeoutException: {err}", err.request) from err
+            raise AvrTimoutError(
+                f"TimeoutException for {err.request.url}: {err}", err.request
+            ) from err
         except httpx.NetworkError as err:
             _LOGGER.debug("Network error exception on request %s: %s", err.request, err)
-            raise AvrNetworkError(f"NetworkError: {err}", err.request) from err
+            raise AvrNetworkError(
+                f"NetworkError for {err.request.url}: {err}", err.request
+            ) from err
         except httpx.RemoteProtocolError as err:
             _LOGGER.debug(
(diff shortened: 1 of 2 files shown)


ol-iver (approved):
Nice, thank you 🙂

PR #391, opened 2026-07-07
Add dedicated mute toggle command

henrikwidlund:
Adds async_mute_toggle to toggle mute state, works over both HTTP and telnet, plus a guard on the regular mute setter.

Relates to #375

### DIFF
--- a/denonavr/volume.py
+++ b/denonavr/volume.py
@@ -350,6 +350,9 @@ class DenonAVRVolume(DenonAVRFoundation):
     async def async_mute(self, mute: bool) -> None:
         """Mute receiver."""
         if mute:
+            if self._muted:
+                return
+
             if self._device.telnet_available:
                 await self._device.telnet_api.async_send_commands(
                     self._device.telnet_commands.command_mute_on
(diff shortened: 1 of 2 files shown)


denonavr/volume.py
```diff
@@ -350,6 +350,9 @@ async def async_set_volume(self, volume: float) -> None:
     async def async_mute(self, mute: bool) -> None:
         """Mute receiver."""
         if mute:
+            if self._muted:
+                return
```
ol-iver: Is there a special reason for this check? 
There is one downside of this check. When you use the HTTP interface, you have to run `async_update` between mute and unmute to update the state. Otherwise, it will do nothing.

denonavr/volume.py
```diff
@@ -359,6 +362,21 @@ async def async_mute(self, mute: bool) -> None:
                     self._device.urls.command_mute_on
                 )
         else:
+            if self._muted is False:
+                return
```
ol-iver: Same here.

ol-iver (commented):
Thanks for your PR 😄

denonavr/volume.py
henrikwidlund: The reason is that you'll get a timeout when using telnet and you send a command with the same value as the one already active on the receiver, it simply doesn't acknowledge the command. I didn't think of the issue with http because where I use the library we have a background thread that calls the update when using http. We can guard this so it only applies to telnet instead.

henrikwidlund: Fixed

denonavr/volume.py
henrikwidlund: Fixed

ol-iver (approved):
Nice 🙂

PR #392, opened 2026-07-07
Correct Bass Sync range to 0-16

henrikwidlund:
The receiver reports and accepts Bass Sync levels in the range 0 to 16, not -10 to 0. Store the reported value as-is (previously negated) and validate the setter against 0-16.

Relates to #375


ol-iver (approved):
Thanks for fixing this bug 😄

PR #393, opened 2026-07-07
Fix volume parsing range and add custom max-volume event

henrikwidlund:
Volume parsing:
- Rewrite convert_volume to parse the telnet 2/3-digit format directly, handle None/invalid lengths, and clamp the result to [-80.0, 18.0] dB so volume can no longer go out of range.
- Simplify _volume_callback to defer parsing to the attr converter.

Custom max-volume event:
- The receiver sends an undocumented MVMAX event that was previously discarded. Propagate it as a dedicated CUST_MAX_VOL (name can be discussed) telnet event and track the receiver's maximum allowed volume (_max_volume, max_volume property).
- Clamp async_set_volume to the max and skip async_volume_up once the max is reached, which otherwise makes the receiver stop responding and time out.

Relates to #375

### DIFF
--- a/denonavr/volume.py
+++ b/denonavr/volume.py
@@ -38,17 +38,40 @@ def convert_muted(value: str) -> bool:
     return bool(value.lower() == STATE_ON)
 
 
-def convert_volume(value: Union[float, str]) -> float:
+def convert_volume(value: str) -> float:
     """Convert volume to float."""
-    if value == "--":
+    if value is None or value == "--":
         return -80.0
-    return float(value)
+
+    value = value.strip()
+    if value.startswith("-"):
+        return float(value)
+
+    if len(value) > 3:
+        _LOGGER.warning(
+            "Volume value length is invalid: %s, defaulting to -80.0 dB", value
+        )
+        return -80.0
+
+    if len(value) < 3:
+        value = -80.0 + float(value)
+    else:
+        value = float(value[:2] + "." + value[2]) - 80.0
+
+    if value < -80.0 or value > 18.0:
+        _LOGGER.warning(
+            "Volume %s converted to %s is out of range. Will use clamping.",
+            value,
+            value,
+        )
+    return max(min(value, 18.0), -80.0)
 
 
 @attr.s(auto_attribs=True, on_setattr=DENON_ATTR_SETATTR)
 class DenonAVRVolume(DenonAVRFoundation):
     """This class implements volume functions of Denon AVR receiver."""
 
+    _max_volume: Optional[float] = attr.ib(default=None)
     _volume: Optional[float] = attr.ib(
         converter=attr.converters.optional(convert_volume), default=None
     )
@@ -83,6 +106,9 @@ class DenonAVRVolume(DenonAVRFoundation):
             self._device.api.add_appcommand_update_tag(tag)
 
         self._device.telnet_api.register_callback("MV", self._volume_callback)
+        self._device.telnet_api.register_callback(
+            "CUST_MAX_VOL", self._max_volume_callback
+        )
         self._device.telnet_api.register_callback("MU", self._mute_callback)
         self._device.telnet_api.register_callback("CV", self._channel_volume_callback)
         self._device.telnet_api.register_callback("PS", self._subwoofer_state_callback)
@@ -92,17 +118,34 @@ class DenonAVRVolume(DenonAVRFoundation):
 
         self._is_setup = True
 
-    def _volume_callback(self, zone: str, event: str, parameter: str) -> None:
+    def _volume_callback(self, zone: str, _event: str, parameter: str) -> None:
         """Handle a volume change event."""
         if self._device.zone != zone:
             return
 
-        if len(parameter) < 3:
-            self._volume = -80.0 + float(parameter)
-        else:
-            whole_number = float(parameter[0:2])
-            fraction = 0.1 * float(parameter[2])
-            self._volume = -80.0 + whole_number + fraction
+        self._volume = parameter
+
+    def _max_volume_callback(self, zone: str, _event: str, parameter: str) -> None:
+        """Handle a max volume change event."""
+        if self._device.zone != zone:
+            _LOGGER.info(
+                "Max volume zone mismatch: expected %s, got %s", self._device.zone, zone
+            )
+            return
+
+        if not parameter.startswith("MAX"):
+            _LOGGER.info("Ignoring invalid max volume parameter: %s", parameter)
+            return
+
+        volume = convert_volume(parameter[4:])
+        if volume == -80.0:
+            # -80.0 indicates error with parsing max volume, minimum allowed is -40.0
+            # Set a reasonable default
+            volume = -20.0
+
+        if self._max_volume != volume:
+            self._max_volume = volume
+            _LOGGER.debug("Set max volume: %s", self._max_volume)
 
     def _mute_callback(self, zone: str, event: str, parameter: str) -> None:
         """Handle a muting change event."""
@@ -227,6 +270,15 @@ class DenonAVRVolume(DenonAVRFoundation):
         return self._volume
 
     @property
+    def max_volume(self) -> float:
+        """
+        Return maximum allowed volume of Denon AVR as float.
+
+        Volume is sent in a format like -50.0.
+        """
+        return self._max_volume if self._max_volume is not None else 18.0
+
+    @property
     def channel_volumes(self) -> Optional[Dict[Channels, float]]:
         """
         Return the channel levels of the device in dB.
@@ -304,6 +356,9 @@ class DenonAVRVolume(DenonAVRFoundation):
     ##########
     async def async_volume_up(self) -> None:
         """Volume up receiver."""
+        if self._max_volume and self._volume and self._volume >= self._max_volume:
+            _LOGGER.debug("Volume already at max value, skipping.")
+            return
         if self._device.telnet_available:
             await self._device.telnet_api.async_send_commands(
                 self._device.telnet_commands.command_volume_up, skip_confirmation=True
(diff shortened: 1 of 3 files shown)


TheEvilCoder42:
I have an AVR-X1700H and ran this branch against it, because I have been working on the same value from a different angle. Posting the measurements in case they are useful: The `async_volume_up` guard and the range clamping both look right to me, but I think `MVMAX` cannot carry `max_volume` on its own.

**Across volume changes it behaves.** Limit held at 60 (−20.0 dB), same volume re-sent three times: `MVMAX 60` each time it was pushed, and `max_volume` read −20.0 throughout. For what #375 asks, not spamming the receiver once the ceiling is reached, that is enough, and it self-corrects on the next `MV`.

**Across limit changes it does not.** Writing the limit over `formiPhoneAppDirect.xml` with a telnet connection open, and reading `SSVCTZMALIM` on a second connection as ground truth:

| write | receiver's limit | this branch's `max_volume` |
|---|---|---|
| `SSVCTZMALIM 70` | −10.0 | −20.0 |
| `SSVCTZMALIM OFF` | no limit | −10.0 |
| `SSVCTZMALIM 80` | 0.0 | −10.0 |
| `SSVCTZMALIM OFF` | no limit | −10.0 |

Two causes: Most limit changes push no `MVMAX` at all, so the value goes stale. And when one is pushed it can be wrong: Clearing a limit of 70 pushed `MVMAX 71`. A `MV?` *query* in that same state answers `MVMAX 98`, the hardware ceiling, so the two disagree. 71 is indistinguishable from a real limit of −9.0 dB, so no guard can filter it.

The first row is the `-80.0 → -20.0` fallback in `_max_volume_callback` firing.
I would drop that branch whatever else happens, it reports a specific limit the receiver never said, and −20.0 is a plausible enough value that it will not look like an error downstream.

Two smaller things:

```python
if self._max_volume and self._volume and self._volume >= self._max_volume:
```

`0.0` is a legal limit (absolute 80, it is one of the three values my zone 2 accepts) and `0.0` is a legal volume, so this guard silently disables itself in both cases. `is not None` on both operands fixes it.

```python
_LOGGER.warning(
    "Volume %s converted to %s is out of range. Will use clamping.",
    value, value,
)
```

`value` has already been reassigned by this point, so both placeholders print the converted number and the original is lost.

**Where the limit does live**, if it helps: `GetAllZoneVolume` already returns `<limit>` per zone and the library parses and discards it, so it is available over HTTP with no new request and no telnet connection. Over telnet the setup-menu event `SSVCTZMALIM` (`SSVCTZ2SLIM` for zone 2) is pushed within 0.1 s of every limit change, including changes made from the front panel or the web GUI. Both were correct in every state I tested.

Being upfront: I have a [branch](https://github.com/TheEvilCoder42/denonavr/tree/max-volume) that reads it that way, with the setter side and zone 2/3 as well, and I have already carried your `async_volume_up` guard into it, including the `is not None` change above, because it is the part of #375 my branch would otherwise leave unsolved.
I would rather land one thing than two competing ones, so before I open anything I would like to know how you want to play it. I am equally happy to rebase mine onto yours, hand you the `SSVCTZMALIM` part to fold into this PR, or open mine separately and have you review it. Your call, you got here first :).

henrikwidlund:
Interesting, I've never seen those on my receiver. I will check your branch and incorporate the missing pieces if that's alright with you?

TheEvilCoder42:
> Interesting, I've never seen those on my receiver. I will check your branch and incorporate the missing pieces if that's alright with you?

Sure, go ahead :)

There's a fair bit in there that's not needed for this fix. Only the `SSVCTZMALIM` handling is relevant here, the rest is setter side and zone 2/3 (groundwork for extending the Home Assistant Integration, I'll open PRs for my other branches).

And since you didn't see those events at all on yours, in case it turns out to be model specific:
`GetAllZoneVolume` already gives you `<limit>` per zone over HTTP, the library just throws it away at the moment.

The three small ones from my comment would be nice to have in there too. The -80.0 to -20.0 fallback is probably worth dropping, the guard could use `is not None`, and I think that warning ends up printing the converted value twice.

Thanks for your efforts!

henrikwidlund:
Okay, I fixed the bugs in the volume parsing and the invalid guards. I tried to get SSVCTZMALIM to produce a Telnet event during startup with different combinations of `?`, but nothing worked, it just timed out waiting for the receiver to acknowledge the command. I ended up adding the `/limit` parameter to the app command, so that sets the value based on its response. Changing the limit on the unit correctly sends the message as well - I simply didn't see that one because I never change the max volume, and didn't look further since the other event did show up.

denonavr/volume.py
```diff
@@ -331,6 +396,14 @@ async def async_set_volume(self, volume: float) -> None:
         Volume is send in a format like -50.0.
         Minimum is -80.0, maximum at 18.0
         """
+        if volume > self.max_volume:
+            _LOGGER.debug(
+                "Volume %s exceeds custom max volume %s. Setting volume to max allowed",
+                volume,
+                self.max_volume,
+            )
+            volume = self.max_volume
```
TheEvilCoder42: If the volume is already at the ceiling, this clamps to the value the receiver
already holds and sends it anyway. The receiver does not answer a write of its
current value, so on the telnet path `_send_lock` stays held for the full 2 s
confirmation timeout.

Measured on an AVR-X1700H, alternating real writes and no-ops, 4 of 4: A write
of the current value gets no reply, a real change echoes immediately.

Maybe `return` here when the clamped value equals `self._volume`?

henrikwidlund: I'll fix this, but I just discovered that my receiver no longer seem to skip ack of commands where the current state can't be changed or already is at the desired value. For example I have my max volume set to -20dB, and even though the current volume is at that level, when I send a volume up command the display on the unit lights up and shows the volume and I get ack immediately. I think this must've changed in the last firmware update.

Clamping if the value can't change is still valid for older units or those who have not updated. Yet another fragmentation in the Denon/Marantz APIs :(

TheEvilCoder42: Ran your current head against an AVR-X1700H with zone 2 powered on. The `/limit` read is correct on this model, zone 2 included:

| zone 2 limit | wire | `max_volume` |
|---|---|---|
| −20.0 | `<limit>-20.0</limit>` | −20.0 |
| −10.0 | `<limit>-10.0</limit>` | −10.0 |
| 0.0 | `<limit>  0.0</limit>` | 0.0 |
| off | `<limit>OFF</limit>` | 18.0 |

Worth noting because `add_zone=True` makes the pattern cover zone 2 and zone 3 as well, and the padded `  0.0` exercises the strip in `convert_max_volume` on live output rather than only on the SR6012 sample. The `SSVCTZ2SLIM` event fires the same way `SSVCTZMALIM` does.

Small thing on c8fef62d6, no action needed unless you want it: Moving the skip inside `if telnet_available` makes sense since the stall is telnet-only, but the `max_volume` clamp moved in with it, so a request above the limit now goes out unclamped on the HTTP path. I checked and the receiver enforces its own ceiling there anyway: `async_set_volume(-10.0)` against a -20.0 limit lands on `MV60`, and `async_volume_up` at the ceiling gets no response, so nothing misbehaves.
Just flagging in case the clamp riding along was unintentional; the effect is a wasted request and the debug line no longer appearing for HTTP users.

henrikwidlund: To be clear, the whole reason why I introduced these checks was because on Telnet you got the timeout when you sent an invalid value or the same value as the one already used on the remote. It was not really intended for rate limiting, that is just a side-bonus. The HTTP commands do not have this issue, and I moved the check inside Telnet as HTTP completely relies on polling and can get out of sync because of that. Sending a few extra requests is better than wrongly ignoring them.

TheEvilCoder42: > Ran your current head against an AVR-X1700H with zone 2 powered on. The
> `/limit` read is correct on this model, zone 2 included:
> 
> | zone 2 limit | wire | `max_volume` |
> |---|---|---|
> | −20.0 | `<limit>-20.0</limit>` | −20.0 |
> | −10.0 | `<limit>-10.0</limit>` | −10.0 |
> | 0.0 | `<limit>  0.0</limit>` | 0.0 |
> | off | `<limit>OFF</limit>` | 18.0 |
> 
> Worth a note because `add_zone=True` makes the pattern cover zone 2 and zone 3
> as well, and the padded `  0.0` exercises the strip in `convert_max_volume` on
> live output rather than only on the SR6012 sample. The `SSVCTZ2SLIM` event
> fires the same way `SSVCTZMALIM` does.

Sorry I mistakenly deleted my comment (it technically didn't belong in this review), I quoted it's relevant content.

My unit sits on the old side of that split. On an AVR-X1700H with the limit at -20.0 and the volume already sitting there, a volume up gets no response at all: nothing echoed, nothing pushed, volume unchanged. So your reading of it looks right. Newer firmware acks these, older firmware stays silent, and keeping the skip is what covers both.

Keeping it to Telnet makes sense too. I hadn't thought about HTTP holding a stale volume between polls. For what it's worth the receiver enforces the ceiling itself on that path anyway: `async_set_volume(-10.0)` against a -20.0 limit lands on `MV60`, so nothing gets through unclamped either way.

The `return` you added in `async_set_volume` is exactly what I was after, so there's nothing outstanding from me on this one. Happy for it to be resolved.

denonavr/volume.py
```diff
@@ -38,17 +38,56 @@ def convert_muted(value: str) -> bool:
...
+
+def convert_volume(value: str) -> float:
     """Convert volume to float."""
-    if value == "--":
+    if value is None or value == "--":
         return -80.0
-    return float(value)
+
+    raw_value = value.strip()
+    if raw_value.startswith("-"):
```
TheEvilCoder42: This treats every value without a `-` as telnet, but HTTP reports 0 dB and above unsigned too: an AVR-X1700H sends Zone2 at 0 dB as `<volume>  0</volume>` and at +18 dB as `<volume> 18</volume>`, which this reads as -80.0 and -62.0 (`main`: 0.0 and 18.0). A plain `float()` here, with the telnet conversion back in `_volume_callback` (line 140) as on `main`, fixes it.

TheEvilCoder42 (changes requested):

PR #394, opened 2026-07-07
Parallelize component setup and update

henrikwidlund:
Run the per-component async setup (soundmode, tonecontrol, additional zones) and the per-component async update (input, soundmode, tonecontrol, volume) concurrently with asyncio.gather instead of sequentially, reducing setup and update latency.

Relates to #375

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


henrikwidlund:
@ol-iver I've dropped support for 3.9 and 3.10 since the testing framework doesn't support the parallel fetches, and they're EOL anyway.

ol-iver (approved):
Nice, the cache still works when we do it this way 🚀 
Thanks 🙂

PR #396, opened 2026-08-20
Add sound mode "DOLBY D+NEURAL:X"

neyzm:
Adds missing sound mode string `DOLBY D+NEURAL:X` under `DOLBY DIGITAL`.

Fixes: home-assistant/core#179518


ol-iver (approved):
Thanks 😄

